PR #1, opened 2023-05-30
List of books in "Livros" struct

cesarFuhr:
In this PR I rewrote the books response as a struct.


server/main.go
```diff
@@ -18,7 +18,26 @@ func getPing(w http.ResponseWriter, r *http.Request) {
 }
 
 func getBooks(w http.ResponseWriter, r *http.Request) {
-	responsebody := `[
+	Livros := []struct {
```
cesarFuhr, line 21: Usually we don't use uppercase letter as the first letter in variable names.

PR #2, opened 2023-06-02
Feat/books as struct valid json

ramon-reichert:
- Maybe (probably) the conflict is because I've change the path of main.go. I decided to discard the "hello word" main.go archive that left on the folder and pull up the "server" main.go from the server folder, just to keep things clean. Can we maintain this chance or should I give a step back?

- The other conflict may be because I actually modified a function that was "ready".

Should I fix it by myself? If yes, how?


Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY go.mod ./
 
 RUN go mod download
 
-COPY server/main.go ./
+COPY main.go ./
```
cesarFuhr, line 11: I don't think you want to do this change, we want the server binary to be in the docker container. Fix it, commit it and push the new commit to the branch.

main.go
```diff
@@ -1,8 +1,47 @@
...
+	}
+
+}
+
+func getBooks(w http.ResponseWriter, r *http.Request) {
+
+	bookslist := []struct { //Struct created to handle data of the list of books
+		Name      string
+		Price     float32
+		Inventory int
```
cesarFuhr: Check on how you can change the names of the properties in the JSON document by adding struct tags.
https://www.digitalocean.com/community/tutorials/how-to-use-struct-tags-in-go

When you have that figured out, change all the property name to lowercase.

cesarFuhr (changes requested):
There are some things you need to change here.

Dockerfile
ramon-reichert: Server/main.go was moved to main.go in "root" directory.

main.go

cesarFuhr (approved):

PR #3, opened 2023-06-13
Feat/create book route

ramon-reichert:
I left the database implementation to another branch so you can parse the code made until now. I will wait for your feedback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM golang:1.20
 
 WORKDIR /app
 
-COPY go.mod ./
+COPY go.mod go.sum ./
 
 RUN go mod download
 
--- a/main.go
+++ b/main.go
@@ -6,22 +6,26 @@ import (
 	"log"
 	"net/http"
 	"strings"
+
+	"github.com/google/uuid"
 )
 
 type Book struct {
-	Name      string  `json:"name"`
-	Price     float32 `json:"price"`
-	Inventory int     `json:"inventory"`
+	Name      string    `json:"name"`
+	Price     float32   `json:"price"`
+	Inventory int       `json:"inventory"`
+	ID        uuid.UUID `json:"id"`
 }
 
 var bookslist = []struct { //Struct created to handle data of the list of books
-	Name      string  `json:"name"`
-	Price     float32 `json:"price"`
-	Inventory int     `json:"inventory"`
+	Name      string    `json:"name"`
+	Price     float32   `json:"price"`
+	Inventory int       `json:"inventory"`
+	ID        uuid.UUID `json:"id"`
 }{
-	{"Book 1", 30.20, 2},
-	{"Book 2", 20.30, 1},
-	{"Book 3", 32.20, 5},
+	{"Book 1", 30.20, 2, uuid.New()},
+	{"Book 2", 20.30, 1, uuid.New()},
+	{"Book 3", 32.20, 5, uuid.New()},
 }
 
 func Ping(w http.ResponseWriter, r *http.Request) {
@@ -51,7 +55,7 @@ func Books(w http.ResponseWriter, r *http.Request) {
 func postBooks(w http.ResponseWriter, r *http.Request) {
 	//TO DO:
 	//Read the Json body
-	var newBook = Book{"sem nome", -1, -1}
+	var newBook = Book{"sem nome", -1, -1, uuid.Nil}
 	decoder := json.NewDecoder(r.Body)
 	err := decoder.Decode(&newBook)
 	if err != nil {
@@ -60,7 +64,7 @@ func postBooks(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	fmt.Printf("newBook: %+v\n", newBook)
-	fmt.Printf("bookslist: %+v", bookslist)
+	fmt.Printf("bookslist: %+v\n", bookslist)
 
 	//Verify if the entry is in a valid format
 	switch {
@@ -80,16 +84,21 @@ func postBooks(w http.ResponseWriter, r *http.Request) {
 	for i := range bookslist {
 		bookAlreadyExists := strings.EqualFold(bookslist[i].Name, newBook.Name)
 		if bookAlreadyExists {
-			warning := fmt.Sprintf("This book already exists in the database: %+v", bookslist[i])
+			warning := fmt.Sprintf("Este livro já existe na base de dados: %+v", bookslist[i])
 			w.Write([]byte(warning))
+			return
 		}
 	}
 
 	//Atribute an ID to the entry
+	newBook.ID = uuid.New()
 
 	//Store the book in the database
+	bookslist = append(bookslist, newBook)
+	fmt.Printf("bookslist: %+v\n", bookslist)
 
 	//Return a sucess message
+	w.Write([]byte("Livro adicionado com sucesso"))
 }
 
 func getBooks(w http.ResponseWriter, r *http.Request) {
--- a/main.go
+++ b/main.go
@@ -50,15 +50,32 @@ func Books(w http.ResponseWriter, r *http.Request) {
 func postBooks(w http.ResponseWriter, r *http.Request) {
 	//TO DO:
 	//Read the Json body
-	var newBook Book
+	var newBook = Book{"sem nome", -1, -1}
 	decoder := json.NewDecoder(r.Body)
 	err := decoder.Decode(&newBook)
 	if err != nil {
 		fmt.Println("error:", err)
-		w.WriteHeader(http.StatusBadRequest) //Verify if the entry is in a valid format
+		w.WriteHeader(http.StatusBadRequest)
+		return
 	}
 	fmt.Printf("newBook: %+v\n", newBook)
 	fmt.Printf("bookslist: %+v", bookslist)
+
+	//Verify if the entry is in a valid format
+	switch {
+	case newBook.Name == "sem nome":
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira um nome para cadastrar o livro."))
+	case newBook.Price == -1: //-1 means that no price was added
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira um preço para cadastrar o livro."))
+	case newBook.Inventory == -1: //-1 menas that no inventory was added
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira a quantidade deste livro em estoque."))
+	default:
+
+	}
+
 	//Verify if that book already exists in the database
 
 	//Atribute an ID to the entry
--- a/main.go
+++ b/main.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 )
 
-func getPing(w http.ResponseWriter, r *http.Request) {
+func Ping(w http.ResponseWriter, r *http.Request) {
 	metodo := r.Method
 	fmt.Printf("metodo: %v\n", metodo)
 	if metodo == "GET" {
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM golang:1.20
 
 WORKDIR /app
 
-COPY go.mod ./
+COPY go.mod go.sum ./
```
cesarFuhr, line 7: Nice catch, you may not have noticed it, but this method actually helps with docker layer caching.

main.go
```diff
@@ -5,9 +5,30 @@ import (
...
+
+	"github.com/google/uuid"
 )
 
-func getPing(w http.ResponseWriter, r *http.Request) {
+type Book struct {
+	Name      string    `json:"name"`
+	Price     float32   `json:"price"`
+	Inventory int       `json:"inventory"`
+	ID        uuid.UUID `json:"id"`
```
cesarFuhr, line 14: Lets move the ID to the first line in the struct definition.

main.go
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
 
 }
 
-func getBooks(w http.ResponseWriter, r *http.Request) {
+func Books(w http.ResponseWriter, r *http.Request) {
+	metodo := r.Method
+	fmt.Printf("metodo: %v\n", metodo)
```
cesarFuhr: I don't think we want this `fmt.Printf` here. We will discuss logging later.

main.go
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
 
 }
 
-func getBooks(w http.ResponseWriter, r *http.Request) {
+func Books(w http.ResponseWriter, r *http.Request) {
+	metodo := r.Method
```
cesarFuhr: Lets not mix languages here, `method` is better.

main.go
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+	fmt.Printf("metodo: %v\n", metodo)
+	switch {
+	case metodo == "GET":
+		getBooks(w, r)
+	case metodo == "POST":
+		postBooks(w, r)
+	default:
+		w.WriteHeader(http.StatusMethodNotAllowed)
+	}
```
cesarFuhr: 1 - Avoid using string literals in comparisons, use the `http` package constants instead.
2 - Instead of a naked switch `switch {`, you can use method as the `switch` target and simplify the syntax. 
3 - You want to return after the dedicated http method function calls. Otherwise (if you had any code after) you would continue execution and it could cause issues. Specially after some time, when you want to tweak this code a little and you already forgot how it actually works.
```suggestion
	switch method {
	case http.MethodGet:
		return getBooks(w, r)
	case http.MethodPost:
		return postBooks(w, r)
	default:
		return w.WriteHeader(http.StatusMethodNotAllowed)
	}

File: main.go [hunk 6]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
-	bookslist := []struct { //Struct created to handle data of the list of books
-		Name      string  `json:"name"`
-		Price     float32 `json:"price"`
-		Inventory int     `json:"inventory"`
-	}{
-		{"Book 1", 30.20, 2},
-		{"Book 2", 20.30, 1},
-		{"Book 3", 32.20, 5},
+func postBooks(w http.ResponseWriter, r *http.Request) {
+	//TO DO:
```
cesarFuhr: Todo?

main.go
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+	//Read the Json body
+	var newBook = Book{"sem nome", -1, -1, uuid.Nil}
```
cesarFuhr: You don't need to manually initialize the values.
This is enough:
```suggestion
	var newBook Book

File: main.go [hunk 8]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+	decoder := json.NewDecoder(r.Body)
+	err := decoder.Decode(&newBook)
```
cesarFuhr: Usually we do it this way:
```suggestion
        err := json.NewDecoder(r.Body).Decode(&newBook)

File: main.go [hunk 9]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		fmt.Println("error:", err)
```
cesarFuhr: Lets change it for:
```suggestion
		log.Println("error:", err)

File: main.go [hunk 10]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	fmt.Printf("newBook: %+v\n", newBook)
+	fmt.Printf("bookslist: %+v\n", bookslist)
```
cesarFuhr: Lets remove this debugging code for now.
```suggestion

File: main.go [hunk 11]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira um nome para cadastrar o livro."))
+	case newBook.Price == -1: //-1 means that no price was added
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira um preço para cadastrar o livro."))
+	case newBook.Inventory == -1: //-1 menas that no inventory was added
+		w.WriteHeader(http.StatusPartialContent)
+		w.Write([]byte("Insira a quantidade deste livro em estoque."))
+	default:
+	}
```
cesarFuhr: In this case you probably want to go with individual `if`s for each check.
This is not partial content, this is a `BadRequest`.
```go
if newBook.Name == "" {
        // do this
}

File: main.go [hunk 12]
```diff
@@ -18,18 +39,69 @@ func getPing(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
 	}
 
+	//Atribute an ID to the entry
+	newBook.ID = uuid.New()
+
+	//Store the book in the database
+	bookslist = append(bookslist, newBook)
+	fmt.Printf("bookslist: %+v\n", bookslist)
```
cesarFuhr: Please remove the debugging code.
```suggestion

**Review by cesarFuhr — CHANGES_REQUESTED** [review]
Really nice, its in the right track.
Please take a look at the comments I made.

Nice job.

File: main.go (hunk 7 above)
Comment (cesarFuhr) [reply main.go]: Go makes sure that all the values are initialized to the zero value for their type on declaration.

File: main.go (hunk 2 above)
Comment (ramon-reichert) [reply main.go:14]: This is really necessary? Because the json.NewDecoder allocates the message received (with name, price and inventory; without ID) in the first 3 fields of the Book struct. I've tried to change the ID to the first line and the NewDecoder get lost.
In the database table I've put the ID in the last field too, even defining it as primary key.

File: main.go (hunk 4 above)
Comment (ramon-reichert) [reply main.go]: Changed. Sometimes eu não me dou conta.

File: main.go (hunk 3 above)
Comment (ramon-reichert) [reply main.go]: Okay. It was like a "flag" to see what was happening.

File: main.go (hunk 2 above)
Comment (cesarFuhr) [reply main.go:14]: Json Marshal getting lost is not a thing. I am sure that it is some other issue, if you are available we can talk today.

File: main.go (hunk 5 above)
Comment (ramon-reichert) [reply main.go]: VScode did not allow me to return a function:
- getBooks(w, r) (no value) used as value
- postBooks(w, r) (no value) used as value
- w.WriteHeader(http.StatusMethodNotAllowed) (no value) used as value

But this way it worked:
switch method {
	case http.MethodGet:
		getBooks(w, r)
		return
	case http.MethodPost:
		postBooks(w, r)
		return
	default:
		w.WriteHeader(http.StatusMethodNotAllowed)
		return
	}

Comment (cesarFuhr) [reply main.go]: That's perfect. It was my mistake. Sorry.

File: main.go (hunk 6 above)
Comment (ramon-reichert) [reply main.go]: Ah, yes. Before start coding, I've made a list:
//TO DO:
//Read the Json body
//Verify if the entry is in a valid format
//Verify if that book already exists in the database
//Atribute an ID to the entry
//Store the book in the database	
//Return a sucess message

I'll have to clean better.

File: main.go (hunk 7 above)
Comment (ramon-reichert) [reply main.go]: Yes, I know. But that is the way I've found of allowing me to check if the request message arrives complete.
It was important specially for the Inventory field, because (depending on the way that we will conceive the user workflow) a 'zero' value could not be a insertion fail, but a intended value. Could we keep this way or it's bad?

File: main.go (hunk 8 above)
Comment (ramon-reichert) [reply main.go]: Ok.

File: main.go (hunk 9 above)
Comment (ramon-reichert) [reply main.go]: ok

File: main.go (hunk 10 above)
Comment (ramon-reichert) [reply main.go]: Ok. Just a doubt: Is it a bad pratice to insert these tests during my work and erase or comment ir after work is done?

File: main.go (hunk 7 above)
Comment (cesarFuhr) [reply main.go]: We can express the absence of a field using a pointer. Maybe we should talk today.

File: main.go (hunk 10 above)
Comment (cesarFuhr) [reply main.go]: It's ok to have debugging code while you are working, whenever you commit code you should clean it up.

File: main.go (hunk 11 above)
Comment (ramon-reichert) [reply main.go]: Ok. I'll find a way of returning all possible warning messages at once.

File: main.go (hunk 2 above)
Comment (ramon-reichert) [reply main.go:14]: Changed. ID is the first.

File: main.go (hunk 11 above)
Comment (ramon-reichert) [reply main.go]: Done.

//Verify if the entry is in a valid format
	blankFields := ""
	if newBook.Name == "" {
		blankFields += "Insira um nome para cadastrar o livro.\n"
	}
	if newBook.Price == nil {
		blankFields += "Insira um preço para cadastrar o livro.\n"
	}
	if newBook.Inventory == nil {
		blankFields += "Insira a quantidade deste livro em estoque.\n"
	}
	if blankFields != "" {
		w.WriteHeader(http.StatusBadRequest)
		w.Write([]byte(blankFields))
		return
	}

Comment (ramon-reichert) [reply main.go]: I,ve changed the order too. 
First: Check if already exists a book with the same name;
Second: Check if all fields are filled.

**Comment by cesarFuhr** [comment]
Hello, why have you closed the PR?

**Comment by ramon-reichert** [comment]
Hello. Yes. I wanted add something before you review. Now is ready.

File: main.go [hunk 13]
```diff
@@ -5,45 +5,114 @@ import (
...
+	}
+
+	//Atribute an ID to the entry
+	newBook.ID = uuid.New()
+
+	//Store the book in the database
+	bookslist = append(bookslist, newBook)
+
+	//Return a sucess message
+	w.Write([]byte("Livro adicionado com sucesso."))
```
cesarFuhr, line 99: Lets return a http 201 "Created", this way we will have a better semantic in the endpoint response.

cesarFuhr (changes requested):
One small thing and we are good to go.

main.go
ramon-reichert: "Created" is the message to the user, correct?
Until now, I kept the messages to the user in portuguese and the code internal messages in english, considering an app usage by brazilians. Should I change everything to english, including this "created" message?

ramon-reichert: Ahhhh. Now I saw it. The http status name is Created!

cesarFuhr: We probably should move everything to English... but lets not stress that now. I am going to create a ticket in trello for you to work on it later.

cesarFuhr (approved):

PR #4, opened 2023-06-23
Feat/database

ramon-reichert:
I know the code is not clean, but its working. I don't know yet how to deal with the database into a container, then I,ve put the code here as draft so you can take a look and we can talk about it.

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,20 +68,25 @@ func postBooks(w http.ResponseWriter, r *http.Request) {
 	}
 
 	//Verify if that book already exists in the database
-	/*for i := range bookslist {
-		bookAlreadyExists := strings.EqualFold(bookslist[i].Name, newBook.Name)
-		if bookAlreadyExists {
-			showBook, err := json.Marshal(bookslist[i])
-			if err != nil {
-				log.Println("error:", err)
-				w.WriteHeader(http.StatusInternalServerError)
-				return
-			}
-			warning := fmt.Sprint("Este livro já existe na base de dados:\n" + string(showBook))
-			w.Write([]byte(warning))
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -1,13 +1,24 @@
...
 
 import (
+	"database/sql"
 	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
-	"strings"
+
+	//"strings"
```
cesarFuhr: Remove this commented import.

cesarFuhr (approved):
Only one small nit pick.
This is great!

Lets chat about how can we make this easier to maintain.
We are starting to get to a point where we need to extract logic and separate concerns.

cesarFuhr:
Lets work together in the container, its easier than you think... we will just need another tool called `docker compose`.

main.go
```diff
@@ -18,7 +27,68 @@ type Book struct {
...
+	case sql.ErrNoRows:
+		unique = true
+		return
+	case nil:
+		unique = false
+		return
+	default:
+		panic(err)
+	}
+}
```
cesarFuhr: Nice job here, usually we avoid using Go naked returns feature, unless we are forced to.
I get that having the returning arguments named is good, but having to track back in the code to see where the returning variables are set is not a good reading experience.

You can have the named arguments and still return the values at the end.
```go
		return true, book

File: main.go [hunk 3]
```diff
@@ -18,7 +27,68 @@ type Book struct {
...
+	case sql.ErrNoRows:
+		fail = true
+		return
+	case nil:
+		fail = false
+		return
+	default:
+		panic(err)
+	}
+}
```
cesarFuhr: Same here about the naked returns.

cesarFuhr (approved):
This is really good work.
Need only a few tweaks, but lets get that done in future iterations.

PR #5, opened 2023-07-14
Feat/error messages

ramon-reichert:
Just two error codes by now.


SQL statments.txt
cesarFuhr, line 1: This will be really useful, please rename the file to `statements.sql`.

error_codes.txt
cesarFuhr, line 1: Just define the errors in the code. There is no need to restate what the code could store. Its just duplicate info that will get out of date soon. Define the errors in an `errors.go` file or something similar, should be easier and better.

```Go
var errResonseCreateBookBadInput = errorResponse{100, "All the fields - name, price and inventory - must be filled correctly."}
var errResonseCreateBookNameConflict = errorResponse{101, "There is already a book with this name on database:"}

File: main.go [hunk 1]
```diff
@@ -25,6 +25,10 @@ type Book struct {
 	Price     *float32  `json:"price"`
 	Inventory *int      `json:"inventory"`
 }
+type errorResponse struct {
+	Error_code    int    `json:"error_code"`
+	Error_message string `json:"error_message"`
+}
```
cesarFuhr: The JSON tag will handle the name of the fields in the JSON document.
Go doesn't use snake case `snake_case` for the names, we use `CamelCase`.
```suggestion

type errorResponse struct {
	Code    int    `json:"error_code"`
	Message string `json:"error_message"`
}

File: main.go [hunk 2]
```diff
@@ -135,51 +143,62 @@ func createBook(w http.ResponseWriter, r *http.Request) {
 	var newBook Book
 	err := json.NewDecoder(r.Body).Decode(&newBook) //Read the Json body and save the entry to newBook
 	if err != nil {
-		log.Println("error while decoding the json entry:", err)
-		w.WriteHeader(http.StatusBadRequest)
+		log.Println(err)
+		w.WriteHeader(http.StatusBadRequest) //SHOULDN'T IT BE STATUS INTERNAL SERVER ERROR?
```
cesarFuhr: Usually not, because this will fail mostly because the JSON is not well formatted. 
Bad formatted JSON is a user error (request error), a bad request.

error_codes.txt
```diff
@@ -0,0 +1,8 @@
+type errorResponse struct {
+	Error_code    int    `json:"error_code"`
+	Error_message string `json:"error_message"`
+}
+
+100, "All the fields - name, price and inventory - must be filled correctly."
+
+101, "There is already a book with this name on database:"
```
cesarFuhr: Whats the reasoning behind starting in 100?

main.go
```diff
@@ -135,51 +143,62 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
-	blankFields := filledFields(newBook) //Verify if all entry fields are filled.
-	if blankFields != "" {
+	filled := filledFields(newBook) //Verify if all entry fields are filled.
+	if !filled {
+		error100, err := json.Marshal(errorResponse{100, "All the fields - name, price and inventory - must be filled correctly."})
```
cesarFuhr: I see you using `json.Marshal` but we usually use `json.NewEncoder(w).Encode(errorResponse)`.
https://pkg.go.dev/encoding/json#Encoder

main.go
```diff
@@ -135,51 +143,62 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
+		error101, err := json.Marshal(errorResponse{101, "There is already a book with this name on database:"})
+		if err != nil {
+			log.Println(err)
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 		w.Header().Set("content-type", "application/json")
 		w.WriteHeader(http.StatusBadRequest)
-		w.Write([]byte(string(sameNameBook)))
+		w.Write(append(error101, sameNameBook...))
```
cesarFuhr: You cannot just append a JSON to another, that's not a valid JSON.

cesarFuhr (changes requested):

main.go
cesarFuhr: Lets return the error response only, without the duplicated book. We can discuss how to improve the error message in a call.

SQL statments.txt
ramon-reichert: Done.

error_codes.txt
ramon-reichert: Done.

main.go

error_codes.txt
ramon-reichert: Just following the pattern in http status code, where the first algarism could indicate a parent class.

main.go
ramon-reichert: Why?

cesarFuhr: Less memory allocation and it makes the code cleaner, there is no need to create a bytes buffer just to write it once you can write directly to the response body with `json.NewEncoder`.

ramon-reichert: Ok

main.go
ramon-reichert: Ok. I left some commented code where the duplicated book is shown until we can talk.

statments.sql
cesarFuhr, line 1: Please fix this file name. The correct word is `statements` not `statments`.

cesarFuhr (approved):
One simple fix and this is good to go.

PR #7, opened 2023-07-18
migrations

ramon-reichert:
It seems to be working, but I have doubts on how to test it.


main.go
```diff
@@ -222,6 +254,10 @@ func main() {
 	dbObject = connectDb()
 	defer dbObject.Close()
 
+	//apply migrations:
+	migration("up")
+	defer migration("down") //how to exit the program to test this???????????
```
cesarFuhr: Let's not migrate down when the service stops, otherwise the database will lose all its data every time we stop the service.

main.go
```diff
@@ -51,6 +55,34 @@ func connectDb() *sql.DB {
 	return db
 }
 
+func migration(upDown string) {
```
cesarFuhr: This is bad design, do not use a string to accumulate functionality in a single function.
Its better to have two functions `migrateDown` and `migrateUp` for example and extract into another function the shared code (setting up the migration connection for example).
They will be simpler, easier to navigate and less complex (with less `if` statements).

cesarFuhr: I get that if the migrations fail we want to stop execution, but its not this function responsibility to stop the service.
Return an error and let the function caller decide what to do with the error.

This is called `error handling`, when a function cannot decide what to do exactly with an error you should give the error more context and return it to the caller.
You can wrap errors using this function:
```go
fmt.Errorf("migrating up: %w", err)
````

https://pkg.go.dev/fmt#Errorf
Wrapping of errors -> https://golangbyexample.com/errorf-function-golang/
https://rollbar.com/blog/golang-wrap-and-unwrap-error/#

Comment (cesarFuhr) [reply main.go]: If you really need to defer a decision to inside the function execution pass a `boolean` or a `enum` that you could do a switch case or a simple `if thisShouldOccur {` check. Using free formed strings is not a good decision because the function caller will have to inspect the function to know what he could pass as a value to it, also since there is no guard rails in the string any typo could cause the function to have a completely unexpected behavior, lets say the user passes `"ups"` what would the migration function do?

**Review by cesarFuhr — CHANGES_REQUESTED** [review]

**Comment by cesarFuhr** [comment]
We will test the migrations when we start writing tests to the database layer.

File: main.go (hunk 1 above)
Comment (ramon-reichert) [reply main.go]: ok

File: main.go (hunk 2 above)
Comment (ramon-reichert) [reply main.go]: ok. In this case, I returned the error to the main function, a it then panics with the wrapped error received.

**Comment by ramon-reichert** [comment]
Should I create another branch based on current main to move on with the other tickets even if this PR is still open?

File: main.go [hunk 3]
```diff
@@ -51,6 +55,26 @@ func connectDb() *sql.DB {
...
 }
 
+func migrationUp() error {
+	driver, err := postgres.WithInstance(dbObject, &postgres.Config{})
+	if err != nil {
+		return fmt.Errorf("migrating up: %w", err)
+	}
+
+	m, err := migrate.NewWithDatabaseInstance(
+		"file://migrations", //HOW IS THE URL INSIDE THE CONTAINER?	WHAT IS RELATIVE/ABSOLUTE PATH?
```
cesarFuhr, line 65: Let's talk about it in a call. It will be easier to show you. 

You copied the migrations to the container in the Dockerfile.

cesarFuhr (approved):
LGTM!

cesarFuhr:
@ramon-reichert Yes you could branch from `main`, but since I already approved merge it and move on.

PR #10, opened 2023-07-25
Feat/get route

ramon-reichert:
Some valid IDs to test:
234fb544-a109-4abc-992c-927813c90ce3
6e2102a8-243f-484a-a97d-131c7f88919a
3911de19-2508-4b27-86d1-e9b77a64fcbb

### DIFF
--- a/database.go
+++ b/database.go
@@ -14,24 +14,24 @@ import (
 //==============DATABASE FUNCTIONS:=================
 
 /* Connects to the database trought a connection string and returns a pointer to a valid DB object (*sql.DB). */
-func connectDb() *sql.DB {
+func connectDb() (*sql.DB, error) {
 
 	db, err := sql.Open("postgres", os.Getenv("DATABASE_URL"))
 	if err != nil {
-		panic(err)
+		return db, fmt.Errorf("connecting to db: %w", err)
 	}
 
 	err = db.Ping()
 	if err != nil {
-		panic(err)
+		return db, fmt.Errorf("connecting to db: %w", err)
 	}
 
 	fmt.Println("Successfully connected!")
-	return db
+	return db, nil
 }
 
 func migrationUp() error {
-	driver, err := postgres.WithInstance(dbObject, &postgres.Config{})
+	driver, err := postgres.WithInstance(dbObjectGlobal, &postgres.Config{})
 	if err != nil {
 		return fmt.Errorf("migrating up: %w", err)
 	}
@@ -51,17 +51,18 @@ func migrationUp() error {
 }
 
 /* Verifies if there is already a book with the name of "newBook" in the database. If yes, returns it. */
-func sameNameOnDB(newBook Book) (unique bool) {
+func sameNameOnDB(newBook Book) (unique bool, unexpected error) {
 	sqlStatement := `SELECT true FROM bookstable WHERE name=$1;`
-	foundRow := dbObject.QueryRow(sqlStatement, newBook.Name)
+	foundRow := dbObjectGlobal.QueryRow(sqlStatement, newBook.Name)
 	var bookAlreadyExists bool
-	switch err := foundRow.Scan(&bookAlreadyExists); err {
+	err := foundRow.Scan(&bookAlreadyExists)
+	switch err {
 	case sql.ErrNoRows:
-		return true
+		return true, nil
 	case nil:
-		return false
+		return false, nil
 	default:
-		panic(err)
+		return false, fmt.Errorf("verifying same name on db: %w", err)
 	}
 }
 
@@ -70,7 +71,7 @@ var errBookNotFound = errors.New("book not found")
 /* Search a book in database based on ID and returns it if succeed. */
 func searchById(id uuid.UUID) (Book, error) {
 	sqlStatement := `SELECT id, name, price, inventory FROM bookstable WHERE id=$1;`
-	foundRow := dbObject.QueryRow(sqlStatement, id)
+	foundRow := dbObjectGlobal.QueryRow(sqlStatement, id)
 	var bookToReturn Book
 	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
 	if err != nil {
@@ -88,19 +89,24 @@ func searchById(id uuid.UUID) (Book, error) {
 }
 
 /* Stores the book into the database, checks and returns it if succeed. */
-func storeOnDB(newBook Book) (fail bool, storedBook Book) {
+func storeOnDB(newBook Book) (Book, error) {
 	sqlStatement := `
 	INSERT INTO bookstable (id, name, price, inventory)
 	VALUES ($1, $2, $3, $4)
 	RETURNING *`
-	createdRow := dbObject.QueryRow(sqlStatement, newBook.ID, newBook.Name, *newBook.Price, *newBook.Inventory)
+	createdRow := dbObjectGlobal.QueryRow(sqlStatement, newBook.ID, newBook.Name, *newBook.Price, *newBook.Inventory)
 	var bookToReturn Book
-	switch err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory); err {
-	case sql.ErrNoRows:
-		return true, Book{}
-	case nil:
-		return false, bookToReturn
-	default:
-		panic(err)
+	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			{
+				return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
+			}
+		default:
+			return Book{}, fmt.Errorf("storing on db: %w", err)
+		}
 	}
+
+	return bookToReturn, nil
 }
--- a/database.go
+++ b/database.go
@@ -100,9 +100,7 @@ func storeOnDB(newBook Book) (Book, error) {
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
-			{
-				return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
-			}
+			return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
 		default:
 			return Book{}, fmt.Errorf("storing on db: %w", err)
 		}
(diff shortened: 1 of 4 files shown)


cesarFuhr:
The build is failing, please check the error here and fix it.
https://github.com/cesarFuhr/books-service/actions/runs/5651778140/job/15310419132

cesarFuhr:
That was nice. Good troubleshooting. : 😁

ramon-reichert:
Bah! I have no idea. Tomorrow I'll try again.

database.go
```diff
@@ -63,6 +64,22 @@ func sameNameOnDB(newBook Book) (unique bool) {
...
+	var bookToReturn Book
+	switch err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory); err {
+	case sql.ErrNoRows:
+		return true, Book{}
+	case nil:
+		return false, bookToReturn
+	default:
+		panic(err)
+	}
+
```
cesarFuhr, line 85: Remove this extra line here.

main.go
```diff
@@ -36,6 +37,38 @@ func ping(w http.ResponseWriter, r *http.Request) {
...
+	id, err := uuid.Parse(justId)
+	if err != nil {
+		log.Println(err)
+		responseJSON(w, http.StatusBadRequest, errIdInvalidFormat)
+		return
+	}
+	//Searching for that ID on database:
+	empty, returnedBook := searchById(id)
+	if empty {
+		w.WriteHeader(http.StatusNoContent) //WE SHOULD DECIDE HOW TO ANSWER THIS. MAYBE A 404(NOTFOUND), BUT THAT WOULD BE A CLIENT ERROR, WHAT I THINK IS NOT THE CASE.
```
cesarFuhr: If the user sends a valid uuid, but we don't have any book with that uuid its a `404`.
If the user sends a invalid uuid, you are catching it in the previous error check.
If we have an unknown issue while trying to find the book in the data base, its a `500`.

database.go
```diff
@@ -63,6 +64,22 @@ func sameNameOnDB(newBook Book) (unique bool) {
 	}
 }
 
+/* Search a book in database based on ID and returns it if succeed. */
+func searchById(id uuid.UUID) (empty bool, storedBook Book) {
```
cesarFuhr: You should return an `error` here,
Wrap it with an `fmt.Errorf` call, add some context data in the message.
Even the `sql.ErrNoRows` can be wrapped.

cesarFuhr: To know exactly what error you returned you can use the `errors.Is` or `error.As`.
https://pkg.go.dev/errors#As
https://pkg.go.dev/errors#Is

cesarFuhr (changes requested):

database.go
```diff
@@ -63,6 +64,22 @@ func sameNameOnDB(newBook Book) (unique bool) {
...
+	sqlStatement := `SELECT id, name, price, inventory FROM bookstable WHERE id=$1;`
+	foundRow := dbObject.QueryRow(sqlStatement, id)
+	var bookToReturn Book
+	switch err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory); err {
+	case sql.ErrNoRows:
+		return true, Book{}
+	case nil:
+		return false, bookToReturn
+	default:
+		panic(err)
```
cesarFuhr: If you return an `error` you don't need to panic here.
Remember when we talked about the function responsibilities and we understood that the migration shouldn't stop execution? This is the same thing.

ramon-reichert: Ok. I did not understand exactly which function may hold the responsibility of stopping execution. It's only the main function? What is the criteria?

database.go
ramon-reichert: I think I'm having a misunderstanding here. Inside the searchById function, we have just the methods QueryRow and Scan, that together seems to return only one possible error, that is ErrNoRows. The way I've written the function, the ErrNoRows is not actually wrapped, but it's treated as the 'empty' bool in line 64. Which else error do I have that I could wrap? I'm not seeing it...

database.go
cesarFuhr: Usually its only `main`. You should avoid panics as much as you can. They cause unexpected behavior. Panics should only happen in a situation where there is no recovery.

Maybe we should talk about it.

database.go
cesarFuhr: The one you calling `panic` on.

ramon-reichert:
I did my try here.

ramon-reichert:
Could be this way too. Seems simpler to me.

cesarFuhr:
Lets talk about it. You need some guidance here, it's simpler than it looks and it's a pattern you will repeat a lot.

ramon-reichert:
Ok. I can talk this morning.

Em sex, 28 de jul de 2023 5:32 AM, César Führ Cará ***@***.***>
escreveu:

database.go
```diff
@@ -63,6 +64,29 @@ func sameNameOnDB(newBook Book) (unique bool) {
...
+func searchById(id uuid.UUID) (Book, error) {
+	sqlStatement := `SELECT id, name, price, inventory FROM bookstable WHERE id=$1;`
+	foundRow := dbObject.QueryRow(sqlStatement, id)
+	var bookToReturn Book
+	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			{
+				bookNotFound = sql.ErrNoRows //Does it resolve the problem of using SQL on main() side or it's just an useless step?
```
cesarFuhr: It does not, use the `errResponse` for now.
Create a new one like `ErrBookNotFound` in the `errors.go` file.

cesarFuhr (changes requested):

database.go
ramon-reichert: I've tried to use errResponse in many ways before, but it is not of error type, that is what I have to return. Furthermore, I thougth that if is not good pratice to use SQL on main() side, in the same way it is not responsibility of database.go handle with http responses. Then I looked for something in the middle of the two options.

cesarFuhr: Fair. Create an error. Pretty simple:
```go
var ErrBookNotFound = errors.New("book not found")
```

Simple, return that when the sql query returns `sql.ErrNoRows`. Than test it in the main execution with `errors.Is(err, ErrBookNotFound)`.

ramon-reichert:
When I have sometime I'll look to the other 'panics'.

database.go
```diff
@@ -88,19 +89,24 @@ func searchById(id uuid.UUID) (Book, error) {
...
-		return false, bookToReturn
-	default:
-		panic(err)
+	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			{
+				return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
+			}
```
cesarFuhr: This should be working fine, but there is not need for the extra brackets.
```suggestion
			return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)

File: main.go [hunk 7]
```diff
@@ -36,6 +39,44 @@ func ping(w http.ResponseWriter, r *http.Request) {
...
+/* Handles a call to /books/(exected id here) and redirects depending on the requested action.  */
+func bookById(w http.ResponseWriter, r *http.Request) {
+	method := r.Method
+	if method == http.MethodGet {
+		getBookById(w, r)
+		return
+	} else {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
```
cesarFuhr: its OK to use the `else` here, but there is no need. I would argue is better not to have it, will make the code simpler and more readable.
```suggestion

	w.WriteHeader(http.StatusMethodNotAllowed)

File: main.go [hunk 8]
```diff
@@ -36,6 +39,44 @@ func ping(w http.ResponseWriter, r *http.Request) {
...
+}
+
+/* Return the book with that specific ID. */
+func getBookById(w http.ResponseWriter, r *http.Request) {
+	//Isolating ID:
+	justId, _ := strings.CutPrefix(r.URL.Path, "/books/")
```
cesarFuhr, line 56: This implementation is a bit fragile. I can show you why, but lets roll with this for now, I don't think it will be a problem for our current use.

main.go
```diff
@@ -104,22 +145,33 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
 
-	fail, storedBook := storeOnDB(newBook) //Store the book in the database
-	if fail {
+	storedBook, err := storeOnDB(newBook) //Store the book in the database
+	if err != nil {
+		if errors.Is(err, errBookNotFound) {
+			log.Println(err)
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
```
cesarFuhr: Does this makes sense? Returning a `not found` when failing to create a book.
You are not passing no ID in the URL, returning `404` to the user states that the path does not exist, but it does, the service only had issues inserting a new book.
So, I would say, lets not care about which error occurred during insertion for now. We may need to check it in the future to make sure we didn't got any conflicts.

cesarFuhr (changes requested):
Only some minor things we need to fix and this should be good to go.

main.go
ramon-reichert: I knew it! I took a while staring at this point when I was writting it. I was wondering that, if I chose 500 instead of 404, checking which error were coming back was almost useless, except for logging it, so I choose 404. Now I see better that it was a mistake. Is it a 500, right?

cesarFuhr: In this case yes, `not found` doesn't make much sense in an insertion. Remove this `error.Is` check altogether. Let it fall into the 500 case.

database.go
```diff
@@ -49,34 +51,60 @@ func migrationUp() error {
...
+	sqlStatement := `SELECT id, name, price, inventory FROM bookstable WHERE id=$1;`
+	foundRow := dbObjectGlobal.QueryRow(sqlStatement, id)
+	var bookToReturn Book
+	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			{
+				return Book{}, fmt.Errorf("searching by ID: %w", errBookNotFound)
+			}
```
cesarFuhr: I am sorry, I let this slip. The brackets are not needed here.
```suggestion
			return Book{}, fmt.Errorf("searching by ID: %w", errBookNotFound)

cesarFuhr (commented):
Almost there, fix the 500 thing we talked about and this little formatting thing I just highlighted.
After that we should be ready to merge.

cesarFuhr (approved):

PR #11, opened 2023-08-02
Feat/list books

ramon-reichert:
I think it may be merged into main after the merge of the other branch(get_route). Since it has only minor fixes to do in that branch, I think it will be fine.


cesarFuhr:
There is something funny with this branch... let me see if I can `rebase` this or fix it somehow before you move on with the changes.

PR #15, opened 2023-08-13
filtering by name and price

ramon-reichert:
I did not touch database_test.go yet. I'll do it if you aprove what I've done until now.

### DIFF
--- a/cmd/api/database_test.go
+++ b/cmd/api/database_test.go
@@ -145,6 +145,24 @@ func TestListBooks(t *testing.T) {
 			is.Equal(returnedBook[0], testBookslist[i])
 		}
 	})
+
+	t.Run("List books without errors filtering by minimum price", func(t *testing.T) {
+		is := is.New(t)
+
+		//Asking all books on the created list with price >= 501
+		returnedBooks, err := listBooks("", 501.00, 9999.99)
+		is.NoErr(err)
+		is.Equal(returnedBooks[:], testBookslist[5:11])
+	})
+
+	t.Run("List books without errors filtering by maximum price", func(t *testing.T) {
+		is := is.New(t)
+
+		//Asking all books on the created list with price <= 501
+		returnedBooks, err := listBooks("", 00.00, 501.00)
+		is.NoErr(err)
+		is.Equal(returnedBooks[:], testBookslist[0:6])
+	})
 }
 
 func toPointer[T any](v T) *T {
--- a/cmd/api/database_test.go
+++ b/cmd/api/database_test.go
@@ -97,40 +97,52 @@ func TestListBooks(t *testing.T) {
 		teardownDB(t)
 	})
 
+	var testBookslist []Book
+	listSize := 11
+
 	t.Run("List books without errors even if there is no books in the database", func(t *testing.T) {
 		is := is.New(t)
 
 		// Write the List Books test here.
-		returnedBooks, err := listBooks()
+		returnedBooks, err := listBooks("", 0.00, 9999.99)
 		is.NoErr(err)
 		is.Equal(returnedBooks, []Book{})
 	})
 
-	t.Run("List books without errors", func(t *testing.T) {
+	t.Run("List all books, no filtering, without errors", func(t *testing.T) {
 		is := is.New(t)
 
 		// Setting up, creating books to be listed.
-		var bookslist []Book
-		listSize := 11
 
 		for i := 0; i < listSize; i++ {
 			b := Book{
 				ID:        uuid.New(),
 				Name:      fmt.Sprintf("Book number %06v", i),
-				Price:     toPointer(float32(40.0)),
-				Inventory: toPointer(10),
+				Price:     toPointer(float32((i * 100) + 1)),
+				Inventory: toPointer(i + 1),
 			}
 
 			newBook, err := storeOnDB(b)
 			is.NoErr(err)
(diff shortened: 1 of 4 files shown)


cmd/api/main.go
```diff
@@ -169,7 +170,50 @@ func createBook(w http.ResponseWriter, r *http.Request) {
 
 /* Returns a list of the stored books. */
 func getBooks(w http.ResponseWriter, r *http.Request) {
-	returnedBooks, err := listBooks()
+
+	err := r.ParseForm() //Verify errors in request query params and populates r.Form with its values.
```
cesarFuhr: This should be used for other types of requests (the ones with the `x-www-form-urlencoded` application type header).
Please use `r.URL.Query()`, it the correct one for this case https://pkg.go.dev/net/url#example-URL.Query.

cesarFuhr: https://go.dev/play/p/HOCbBF8c9UI

cmd/api/main.go
```diff
@@ -169,7 +170,50 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Println(err)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	//Adapt query params:
+	name := r.FormValue("name")
+	if name == "" {
+		name = "%"
```
cesarFuhr: This is leaking SQL logic into the request handling layer, `%` is SQL specific and not `http` related.
Its better to pass the empty string to the database function and let it decide what that should be in the SQL statement.

cmd/api/main.go
```diff
@@ -169,7 +170,50 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var minPrice32 float32
+	minPriceStr := r.FormValue("min_price")
+	if minPriceStr != "" {
+		minPrice64, err := strconv.ParseFloat(minPriceStr, 32)
+		if err != nil {
+			log.Println(err)
+			w.WriteHeader(http.StatusInternalServerError)
```
cesarFuhr: This is probably a user error, if the user passes an invalid float string this error would occur.

cmd/api/main.go
```diff
@@ -169,7 +170,50 @@ func createBook(w http.ResponseWriter, r *http.Request) {
...
+		minPrice32 = 0
+	}
+
+	var maxPrice32 float32
+	maxPriceStr := r.FormValue("max_price")
+	if maxPriceStr != "" {
+		maxPrice64, err := strconv.ParseFloat(maxPriceStr, 32)
+		if err != nil {
+			log.Println(err)
+			w.WriteHeader(http.StatusInternalServerError)
```
cesarFuhr: Same here.

cesarFuhr (changes requested):
After these changes please add tests.
We will improve it more after the tests are written, this a good development flow:
1 - Make it work
2 - Write the tests to make sure its working as expected
3 - Refactor the code to make it easier to read/faster

ramon-reichert:
Sorry, I've forgotten to write the test filtering by price. I'll do it now.

cmd/api/database_test.go
```diff
@@ -97,40 +97,71 @@ func TestListBooks(t *testing.T) {
...
+	})
+
+	t.Run("List books without errors filtering by maximum price", func(t *testing.T) {
+		is := is.New(t)
+
+		//Asking all books on the created list with price <= 501
+		returnedBooks, err := listBooks("", 00.00, 501.00)
 		is.NoErr(err)
-		is.Equal(returnedBooks, bookslist)
+		is.Equal(returnedBooks[:], testBookslist[0:6])
```
```suggestion
		is.Equal(returnedBooks, testBookslist[0:6])

File: cmd/api/database_test.go [hunk 6]
```diff
@@ -97,40 +97,71 @@ func TestListBooks(t *testing.T) {
...
+		}
+	})
+
+	t.Run("List books without errors filtering by minimum price", func(t *testing.T) {
+		is := is.New(t)
+
+		//Asking all books on the created list with price >= 501
+		returnedBooks, err := listBooks("", 501.00, 9999.99)
+		is.NoErr(err)
+		is.Equal(returnedBooks[:], testBookslist[5:11])
```
```suggestion
		is.Equal(returnedBooks, testBookslist[5:11])

File: cmd/api/database_test.go [hunk 7]
```diff
@@ -97,40 +97,71 @@ func TestListBooks(t *testing.T) {
...
+				Price:     toPointer(float32((i * 100) + 1)),
+				Inventory: toPointer(i + 1),
 			}
 
 			newBook, err := storeOnDB(b)
 			is.NoErr(err)
 			is.Equal(newBook, b)
-			bookslist = append(bookslist, b)
+			testBookslist = append(testBookslist, b)
 		}
```
cesarFuhr: All the following tests depend on this `for` loop that creates the books to be listed.
This means that the other tests depend on this test to run correctly, lets move this `for` loop to line `111`, that way you set up all the tests in one spot, outside of the `t.Run` function.

cesarFuhr (approved):
Other than those little changes, LGTM.

cesarFuhr (commented):
There is only one last test missing, filter by name with partial match.

Lets say you have a book called `Book of Frogs` in the database. If you filter the search based on `name`, with the filter value `Frog`, this book should be returned.

PR #16, opened 2023-08-21
Feat/list ordering

ramon-reichert:
Some decisions we need to take. I'll try to adress it with comments in the PR.

### DIFF
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -4,7 +4,6 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"log"
 	"os"
 
 	"github.com/golang-migrate/migrate/v4"
@@ -75,7 +74,7 @@ func searchById(id uuid.UUID) (Book, error) {
 	sqlStatement := `SELECT id, name, price, inventory FROM bookstable WHERE id=$1;`
 	foundRow := dbObjectGlobal.QueryRow(sqlStatement, id)
 	var bookToReturn Book
-	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
@@ -94,13 +93,11 @@ func listBooks(name string, minPrice32, maxPrice32 float32, sortBy string) ([]Bo
 		name = "%"
 	}
 
-	log.Println(sortBy) //DELETE THIS DEBUG CODE
-
-	sqlStatement :=
-		`SELECT * FROM bookstable 
+	sqlStatement := fmt.Sprint(`SELECT * FROM bookstable 
 	WHERE name LIKE $1
 	AND price BETWEEN $2 AND $3	
-	ORDER BY name ASC;`
+	ORDER BY `, sortBy, ` ASC;`)
+
 	rows, err := dbObjectGlobal.Query(sqlStatement, name, minPrice32, maxPrice32)
 	if err != nil {
 		return nil, fmt.Errorf("listing books from db: %w", err)
@@ -109,7 +106,7 @@ func listBooks(name string, minPrice32, maxPrice32 float32, sortBy string) ([]Bo
 	bookslist := []Book{}
 	var bookToReturn Book
 	for rows.Next() {
-		err = rows.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+		err = rows.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
 		if err != nil {
 			return nil, fmt.Errorf("listing books from db: %w", err)
 		}
@@ -133,7 +130,7 @@ func storeOnDB(newBook Book) (Book, error) {
 	RETURNING *`
 	createdRow := dbObjectGlobal.QueryRow(sqlStatement, newBook.ID, newBook.Name, *newBook.Price, *newBook.Inventory)
 	var bookToReturn Book
-	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory)
+	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -92,9 +92,11 @@ func listBooks(name string, minPrice32, maxPrice32 float32, sortBy, sortDirectio
 	if name == "" {
 		name = "%"
 	}
+	name = fmt.Sprint("%", name, "%")
+	fmt.Println(name)
 
 	sqlStatement := fmt.Sprint(`SELECT * FROM bookstable 
-	WHERE name LIKE $1
+	WHERE name ILIKE $1
 	AND price BETWEEN $2 AND $3	
 	ORDER BY `, sortBy, ` `, sortDirection, ` ;`)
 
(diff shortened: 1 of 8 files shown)


cmd/api/main.go
```diff
@@ -204,8 +205,31 @@ func getBooks(w http.ResponseWriter, r *http.Request) {
...
+		break
+	case "updated_at":
+		break //IMPLEMENT THIS LATER, WHIT FUNCTION UPDATE BOOK
+		//https://x-team.com/blog/automatic-timestamps-with-postgresql/
+		//https://www.postgresqltutorial.com/postgresql-date-functions/postgresql-current_timestamp/
+		//https://www.postgresql.org/docs/15/sql-createtrigger.html
+	default:
+		responseJSON(w, http.StatusBadRequest, errResponseQuerySortByInvalid)
+		return
+	}
```
cesarFuhr: This is good, but lets extract it to a dedicated function.

migrations/000001_bookstable.up.sql
```diff
@@ -1,7 +1,26 @@
...
 CREATE TABLE IF NOT EXISTS public.bookstable
 (
     id uuid PRIMARY KEY NOT NULL,
     name text   UNIQUE NOT NULL,
     price numeric(6,2),
-    inventory integer
-)
\ No newline at end of file
+    inventory integer,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
ramon-reichert: Should we let Postgresql to deal with the date/time data or our api should create this data and just store it on db?

migrations/000001_bookstable.up.sql
cesarFuhr, line 1: You shouldn't update migrations after they are already applied.
To update the table structure, create a new `000002_add_timestamps_to_books_table.up.sql` and use `ALTER TABLE`.
Don't forget to also have a `.down.sql` file to revert the changes.

cmd/api/main.go
```diff
@@ -21,6 +21,8 @@ type Book struct {
 	Name      string    `json:"name"`
 	Price     *float32  `json:"price"`
 	Inventory *int      `json:"inventory"`
+	CreatedAt string    `json:"created_at"`
```
ramon-reichert: Do we want to use/show the information of date/time inside our api code or it is just do order the list that db returns?
In other words: Should we created fields for it in our Book struct or not? This will change the way we make tests.

cmd/api/database_test.go
ramon-reichert, line 1: I'm already facing the problem here on line 46, for example.

t.Run("creates a book without errors", func(t *testing.T) {
		is := is.New(t)
		b := Book{
			ID:        uuid.New(),
			Name:      "A new book`",
			Price:     toPointer(float32(40.0)),
			Inventory: toPointer(10),
		}

migrations/000001_bookstable.up.sql
cesarFuhr: Either is fine. There is good and bad things in both.
We just need to be consistent, if you are doing it in the Go code always do it that way, if the queries are taking care of it, always do it that way.
I like having it done in the Go side, it makes you not depend as much in the database functionality.

migrations/000001_bookstable.up.sql
ramon-reichert: Even in this stage of development? What I've done is to kill all data on db container with a docker compose down.

cmd/api/main.go
cesarFuhr: I would say its OK to show the timestamps on the API, even good to have it.
Lets have it in the API.

cesarFuhr: That said, you could just hide it in the JSON data with an explicit json ignore tag, like this `json:"-"`.

cmd/api/database_test.go
cesarFuhr: Lets expose it in the API.

migrations/000001_bookstable.up.sql
cesarFuhr: Even in the development stage, if you ever want to get the system back to a older state you will have issues to apply the changed migrations. Migrations files give you the ability to move your system back in time without destroying data.

Who would do those manual operations in the database of our deployed system?
We want to manage the database/system as automatically as we can.

migrations/000001_bookstable.up.sql
ramon-reichert: I think that doing it in API will be easier, too. There is some preferable function to write the timestamp or should I discover it on my own?

cesarFuhr: Use `go` `time.Time` type. SQL driver should be smart enough to know how to convert that.
Just make sure you are always working in the UTC timezone when comparing/read/writing timestamps.

migrations/000001_bookstable.up.sql
ramon-reichert: Ok. About the books that are already there (on fly), when we add the new columns to the table, should we:
1) Drop them all manually and start from zero with the new table?
2) Just let the new column fields blank?
3) Give to them a default value to fill the new column fields?

cesarFuhr: Lets go with give them a default `NOW()` value.
I know its not precise, because the books are already created... but its not an issue for us.

cesarFuhr: If you don't want to give the column a default, you can do an `UPDATE` SQL query on all the existing table rows setting their `created_at` and `updated_at` columns to `NOW()`, in the migration `000002` file after you created the new books table columns.

cmd/api/main.go
```diff
@@ -215,6 +227,31 @@ func getBooks(w http.ResponseWriter, r *http.Request) {
 	responseJSON(w, http.StatusOK, returnedBooks)
 }
 
+func extractOrderParmams(query url.Values) (sortBy string, valid bool) {
```
cesarFuhr:
```suggestion
func extractOrderParams(query url.Values) (sortBy string, valid bool) {

File: cmd/api/main.go [hunk 5]
```diff
@@ -21,6 +23,8 @@ type Book struct {
 	Name      string    `json:"name"`
 	Price     *float32  `json:"price"`
 	Inventory *int      `json:"inventory"`
+	CreatedAt time.Time `json:"-"`
+	UpdatedAt time.Time `json:"-"`
```
cesarFuhr, line 27: You decided not to expose it in the JSON response?

cesarFuhr:
Go `time.Time` type cannot be directly compared, it has a dedicated method to compare two time values `sometime.Equal(othertime)`. I also forced Postgres to use `UTC` as its default timezone now.
Take a look at this commit: https://github.com/cesarFuhr/books-service/pull/16/commits/ce1c0c26f4903e0366b7ba1d3bdf4c3a30b5913f. Let me know if you have any questions.

cmd/api/database.go
```diff
@@ -88,16 +88,18 @@ func searchById(id uuid.UUID) (Book, error) {
 }
 
 /* Returns filtered content of database in a list of books*/
-func listBooks(name string, minPrice32, maxPrice32 float32) ([]Book, error) {
+func listBooks(name string, minPrice32, maxPrice32 float32, sortBy, sortDirection string) ([]Book, error) {
 	if name == "" {
 		name = "%"
 	}
+	name = fmt.Sprint("%", name, "%")
+	fmt.Println(name)
```
cesarFuhr: Is this debug code?
If it is, please remove it.

cmd/api/database.go
```diff
@@ -88,16 +88,18 @@ func searchById(id uuid.UUID) (Book, error) {
 }
 
 /* Returns filtered content of database in a list of books*/
-func listBooks(name string, minPrice32, maxPrice32 float32) ([]Book, error) {
+func listBooks(name string, minPrice32, maxPrice32 float32, sortBy, sortDirection string) ([]Book, error) {
 	if name == "" {
 		name = "%"
 	}
+	name = fmt.Sprint("%", name, "%")
```
cesarFuhr: If `name` is empty you will end up with this:
```go
name = fmt.Sprint("%", "%", "%")
```
Does this works as expected? seems a little wrong to have triple `%` though... you could change it to the following and avoid that case.

```suggestion
	if name != "" {
		name = fmt.Sprint("%", name, "%")
	} else {
	        name = "%"
        }

cmd/api/main.go
cesarFuhr: @ramon-reichert any answer to this question?

cesarFuhr (approved):

cesarFuhr:
Please fix the little things left, other than that. LGTM.

cmd/api/main.go
ramon-reichert: As your suggestion:
"[cesarFuhr](https://github.com/cesarFuhr) [last week](https://github.com/cesarFuhr/books-service/pull/16#discussion_r1300043718)
That said, you could just hide it in the JSON data with an explicit json ignore tag, like this json:"-".

cmd/api/database.go
ramon-reichert: Yes. Sorry

cmd/api/database.go
ramon-reichert: It works, but your way is better.

cmd/api/main.go
cesarFuhr: Please don't forget to fix this typo.

PR #18, opened 2023-08-28
Feat/update route

ramon-reichert:
Recreated.

### DIFF
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -115,12 +115,7 @@ func storeOnDB(bookEntry Book) (Book, error) {
 	var bookToReturn Book
 	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
 	if err != nil {
-		switch err {
-		case sql.ErrNoRows:
-			return Book{}, fmt.Errorf("storing on db: %w", errResponseBookNotFound)
-		default:
-			return Book{}, fmt.Errorf("storing on db: %w", err)
-		}
+		return Book{}, fmt.Errorf("storing on db: %w", err)
 	}
 
 	return bookToReturn, nil
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"database/sql"
-	"errors"
 	"fmt"
 	"os"
 
@@ -51,24 +50,6 @@ func migrationUp() error {
 	return nil
 }
 
-/* Verifies if there is already a book with the name of bookEntry in the database. If yes, returns it. */
-func sameNameOnDB(bookEntry Book) (unique bool, unexpected error) {
-	sqlStatement := `SELECT true FROM bookstable WHERE name=$1;`
-	foundRow := dbObjectGlobal.QueryRow(sqlStatement, bookEntry.Name)
-	var bookAlreadyExists bool
-	err := foundRow.Scan(&bookAlreadyExists)
-	switch err {
-	case sql.ErrNoRows:
-		return true, nil
-	case nil:
-		return false, nil
-	default:
-		return false, fmt.Errorf("verifying same name on db: %w", err)
-	}
-}
-
-var errBookNotFound = errors.New("book not found")
-
 /* Searches a book in database based on ID and returns it if succeed. */
 func searchById(id uuid.UUID) (Book, error) {
 	sqlStatement := `SELECT id, name, price, inventory, created_at, updated_at FROM bookstable WHERE id=$1;`
@@ -78,7 +59,7 @@ func searchById(id uuid.UUID) (Book, error) {
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
-			return Book{}, fmt.Errorf("searching by ID: %w", errBookNotFound)
+			return Book{}, fmt.Errorf("searching by ID: %w", errResponseBookNotFound)
 		default:
 			return Book{}, fmt.Errorf("searching by ID: %w", err)
 		}
@@ -136,7 +117,7 @@ func storeOnDB(bookEntry Book) (Book, error) {
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
-			return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
+			return Book{}, fmt.Errorf("storing on db: %w", errResponseBookNotFound)
 		default:
 			return Book{}, fmt.Errorf("storing on db: %w", err)
 		}
@@ -158,7 +139,7 @@ func updateOnDB(bookEntry Book) (Book, error) {
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
-			return Book{}, fmt.Errorf("updating on db: %w", errBookNotFound)
+			return Book{}, fmt.Errorf("updating on db: %w", errResponseBookNotFound)
 		default:
 			return Book{}, fmt.Errorf("updating on db: %w", err)
 		}
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -51,10 +51,10 @@ func migrationUp() error {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


cmd/api/database_test.go
```diff
@@ -59,6 +59,40 @@ func TestCreateBook(t *testing.T) {
...
+		//Updating the created book.
+		b.Name = "The book is now updated"
+		b.Price = toPointer(float32(50.0))
+		b.Inventory = toPointer(9)
+		b.UpdatedAt = time.Now().UTC().Round(time.Millisecond)
+
+		updatedBook, err := updateOnDB(b)
+		is.NoErr(err)
+		compareBooks(is, updatedBook, b)
+	})
```
cesarFuhr, line 93: Nice test. Lets add some more tests to cover the error cases, at least the Book not found error.

cmd/api/main.go
```diff
@@ -97,17 +110,17 @@ func books(w http.ResponseWriter, r *http.Request) {
 }
 
 /* Verifies if all entry fields are filled and returns a warning message if so. */
-func filledFields(newBook Book) bool {
+func filledFields(bookEntry Book) bool {
```
cesarFuhr: Lets start some refactor here.
There is not real reason to keep this function returning a boolean.
In any of the empty fields conditions occur, just return the `errResponseBookEntryBlankFileds` error.
It will read much nicer and, eventually if we want to, we can make the function return a wide variety of errors.

cmd/api/main.go
```diff
@@ -127,45 +140,106 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
-	filled := filledFields(newBook) //Verify if all entry fields are filled.
+	filled := filledFields(bookEntry) //Verify if all entry fields are filled.
 	if !filled {
-		responseJSON(w, http.StatusBadRequest, errResponseCreateBookBlankFileds)
+		responseJSON(w, http.StatusBadRequest, errResponseBookEntryBlankFileds)
+		return false, bookEntry
+	}
+
+	return true, bookEntry
+}
```
cesarFuhr: I get what you are trying here, but I don't think this approach is good.
Lets have these two responsibilities separated, one function to validate the data and another one to render the error in the response.
Our errors already carry a lot of detail, you can use them alone to render the correct error messages.

cmd/api/main.go
```diff
@@ -127,45 +140,106 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
+			log.Println(err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if !unique {
+			responseJSON(w, http.StatusBadRequest, errResponseBookEntryNameConflict)
+			return
+		}
+	}
```
cesarFuhr: I am not feeling good about this, if the user can run into conflicts and we consider `name` as an unique table column maybe we shouldn't let the user change the Books name? I am still not convinced that we should prevent the user from creating books with the same name, we don't use `name` anymore as an identifier anyway.

ramon-reichert:
missing:

add more tests;
remove unique constrain from column 'name'
DOUBT: Why the `put` cannot change just on field, if needed? Inventory, for example.

cmd/api/database.go
```diff
@@ -125,22 +106,44 @@ func listBooks(name string, minPrice32, maxPrice32 float32, sortBy, sortDirectio
...
 	RETURNING *`
-	createdRow := dbObjectGlobal.QueryRow(sqlStatement, newBook.ID, newBook.Name, *newBook.Price, *newBook.Inventory, newBook.CreatedAt, newBook.UpdatedAt)
+	createdRow := dbObjectGlobal.QueryRow(sqlStatement, bookEntry.ID, bookEntry.Name, *bookEntry.Price, *bookEntry.Inventory, bookEntry.CreatedAt, bookEntry.UpdatedAt)
 	var bookToReturn Book
 	err := createdRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
-			return Book{}, fmt.Errorf("storing on db: %w", errBookNotFound)
+			return Book{}, fmt.Errorf("storing on db: %w", errResponseBookNotFound)
```
cesarFuhr: Don't think this will be an issue, but conceptually this is wrong.
How can a `Create Book` operation fail with a `Book not found` error? Looks like this should be a 500 instead.

cmd/api/main.go
```diff
@@ -127,45 +136,75 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
+			Message: errResponseBookEntryInvalidJSON.Message + err.Error(),
 		}
 		responseJSON(w, http.StatusBadRequest, errR)
 		return
 	}
 
-	filled := filledFields(newBook) //Verify if all entry fields are filled.
-	if !filled {
-		responseJSON(w, http.StatusBadRequest, errResponseCreateBookBlankFileds)
+	fillErr := filledFields(bookEntry) //Verify if all entry fields are filled.
```
cesarFuhr: Its ok to name your errors, but usually when there is only one error being handled we just use the `err` name.
If you get into more complex cases where you have more than one error in the scope/handling then its more usual to give error variables custom names.

cmd/api/main.go
```diff
@@ -127,45 +136,75 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if !unique {
-		responseJSON(w, http.StatusBadRequest, errResponseCreateBookNameConflict)
+	responseJSON(w, http.StatusOK, updatedBook)
+}
+
+/* Stores the entry as a new book in the database, if there isn't one with the same name yet. */
```
cesarFuhr: Since we dropped the `name` unique constraint, this comment is outdated.

cmd/api/main.go
```diff
@@ -127,45 +136,75 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
+		log.Println(err)
+		errR := errResponse{
+			Code:    errResponseBookEntryInvalidJSON.Code,
+			Message: errResponseBookEntryInvalidJSON.Message + err.Error(),
+		}
+		responseJSON(w, http.StatusBadRequest, errR)
+		return
+	}
+
+	fillErr := filledFields(bookEntry) //Verify if all entry fields are filled.
```
cesarFuhr: Same as in the update method, usually we just use `err`.

cesarFuhr (commented):
I noticed some little things, fix them and we can move forward.

cmd/api/database.go
ramon-reichert: You're right. The book the was supposed to be created was not found. In other words: was not created. Could be errBookNotCreated. Buuut, we named it as errBookNotFound because it is an ErrNoRows, and we have other case to unexpected errors when the book is not created either. Furthermore, we are already returning a 500 for both cases.
I see that the big mistake here is to replace errBookNotFound (an internal thing) by an errResponse, that is aiming at a http RESPONSE. Could we just go back to errBookNotFound?

cmd/api/main.go
ramon-reichert: Ok, I'll change it.

cmd/api/database.go
cesarFuhr: Just wrap the error and return it. Our handler will take care of making it a 500.

ramon-reichert: What should I maintain: errBookNotFound or errResponseBookNotFound?

ramon-reichert: Or neither one? We could take out the switch and just return a wrapped err.

cesarFuhr: Yep, there is no need to check for a specific error. Any error will cause a 500.

cesarFuhr:
> missing:
> 
> add more tests; remove unique constrain from column 'name' DOUBT: Why the `put` cannot change just on field, if needed? Inventory, for example.

About the PUT question. The convention around the PUT method is that you will overwrite the whole entity, for partial uploads usually we use the PATCH method.

cesarFuhr (approved):

PR #19, opened 2023-09-09
Feat/archive route

ramon-reichert:
From this point, its easy to create a function to restore the archived books, if you want. Just define the http method for it, if so.

### DIFF
--- a/cmd/api/database_test.go
+++ b/cmd/api/database_test.go
@@ -211,7 +211,7 @@ func TestListBooks(t *testing.T) {
 		is := is.New(t)
 
 		// Write the List Books test here.
-		returnedBooks, err := listBooks("", 0.00, 9999.99, "name", "asc")
+		returnedBooks, err := listBooks("", 0.00, 9999.99, "name", "asc", true)
 		is.NoErr(err)
 		is.Equal(returnedBooks, []Book{})
 	})
@@ -237,7 +237,7 @@ func TestListBooks(t *testing.T) {
 		is := is.New(t)
 
 		//Asking all books on the list
-		returnedBooks, err := listBooks("", 0.00, 9999.99, "name", "asc")
+		returnedBooks, err := listBooks("", 0.00, 9999.99, "name", "asc", true)
 		is.NoErr(err)
 		for i, expected := range testBookslist {
 			compareBooks(is, returnedBooks[i], expected)
@@ -249,7 +249,7 @@ func TestListBooks(t *testing.T) {
 
 		// Testing, by name, each book on the created list.
 		for i := 0; i < listSize; i++ {
-			returnedBook, err := listBooks(fmt.Sprintf("Book number %06v", i), 0.00, 9999.99, "name", "asc")
+			returnedBook, err := listBooks(fmt.Sprintf("Book number %06v", i), 0.00, 9999.99, "name", "asc", true)
 			is.NoErr(err)
 			is.True(len(returnedBook) == 1)
 			compareBooks(is, returnedBook[0], testBookslist[i])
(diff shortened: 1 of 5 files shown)


ramon-reichert:
This repetead commits is because I've forgotten a debug code, tried to undo a commit, and vscode get lost. My mistake.

cesarFuhr:
> This repetead commits is because I've forgotten a debug code, tried to undo a commit, and vscode get lost. My mistake.

Ok, no problem.
If you are curious on how we could clean the git history, let me know.

cmd/api/database_test.go
```diff
@@ -58,7 +58,57 @@ func TestCreateBook(t *testing.T) {
...
+		is.NoErr(err)
+		compareBooks(is, newBook, b)
+
+		//Changing the status of 'arquived' field of local book to be compare afterwards.
+		b.Archived = true
+
+		//Archiving the created book.
+		archivedBook, err := archiveStatusOnDB(b.ID, true)
+		is.NoErr(err)
+		compareBooks(is, archivedBook, b)
```
```suggestion
		//Archiving the created book.
		archivedBook, err := archiveStatusOnDB(b.ID, true)
		is.NoErr(err)
		
		//Changing the status of 'arquived' field of local book to be compare afterwards.
		b.Archived = true
		compareBooks(is, archivedBook, b)

File: cmd/api/database_test.go [hunk 2]
```diff
@@ -262,14 +312,39 @@ func TestListBooks(t *testing.T) {
...
+		returnedBook, err := listBooks("", 0.00, 9999.99, "name", "asc", false)
+		is.NoErr(err)
+		is.True(len(returnedBook) == (listSize - 1))
+
+		for i := 0; i < (listSize - 1); i++ {
+			is.True(returnedBook[i].Archived == false)
+		}
+	})
+
+	t.Run("List an archived book should return an empty array of books, no errors.", func(t *testing.T) {
```
```suggestion
t.Run("Filtering a list by an archived book name returns an empty list, no errors.", func(t *testing.T) {

File: cmd/api/main.go [hunk 3]
```diff
@@ -136,7 +140,29 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
...
 }
 
-/* Verifies with all fields are correctly filled and update the book in the db. */
+/* Set 'true' to column 'archived' in the database */
+func archiveStatusBook(w http.ResponseWriter, r *http.Request) {
+	id, err := isolateId(w, r)
+	if err != nil {
+		return
+	}
+	archived := true //THIS FUNCTION IS READY TO BE IMPROVED TO ALLOW RESTORING BOOKS FROM ARCHIVE. JUST FIX THE ROUTING!
```
cesarFuhr: Can you explain this comment?

cmd/api/main.go
```diff
@@ -136,7 +140,29 @@ func responseJSON(w http.ResponseWriter, status int, body any) {
 	}
 }
 
-/* Verifies with all fields are correctly filled and update the book in the db. */
+/* Set 'true' to column 'archived' in the database */
```
cesarFuhr: This does not look as a HTTP handler comment, looks more like a database method comment.
This handler should get more business oriented comments.

migrations/000004_add_archive_to_books_table.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.bookstable
+  ADD COLUMN IF NOT EXISTS archived boolean DEFAULT false;
```
cesarFuhr:
```suggestion
  ADD COLUMN IF NOT EXISTS archived BOOLEAN DEFAULT false;

cmd/api/database_test.go
ramon-reichert: I did not understand the reason for that inversion. The behavior seems to be the same.

cmd/api/main.go
ramon-reichert: Sure! On trello:
"Should we ofer a way of re-activate an archived book?
Lets not worry about it now, create a ticket here in Trello so we can add this functionality later."

I was not thinking on the "re-activate" tool, but, looking at the code I've made for "archiving" tool, I've noticed that it will be pretty easy to give a few more steps and let the whole thing ready. So I commented to don't forget it.

ramon-reichert: Buuut, to allow me to continue, I would need your definition of the proper http method responsible for DEarchivate books.

cmd/api/database_test.go
cesarFuhr: It is the same, but it's preferable to set/create variables closer to where they are used. Helps to create the code narrative.

cmd/api/main.go
cesarFuhr: We would probably want to create a route for that. It's not a functionality that should be used often and it's probably an "admin" use case.

cmd/api/database_test.go
ramon-reichert: ok

cesarFuhr (approved):

PR #20, opened 2023-09-10
down migrations tested

ramon-reichert:
I've tested the down migrations manually on pgadmin, one by one, and its working fine.
And now we have this automated test to general checking.

### DIFF
--- a/cmd/api/database.go
+++ b/cmd/api/database.go
@@ -52,10 +52,10 @@ func migrationUp() error {
 
 /* Searches a book in database based on ID and returns it if succeed. */
 func searchById(id uuid.UUID) (Book, error) {
-	sqlStatement := `SELECT id, name, price, inventory, created_at, updated_at FROM bookstable WHERE id=$1;`
+	sqlStatement := `SELECT id, name, price, inventory, created_at, updated_at, archived FROM bookstable WHERE id=$1;`
 	foundRow := dbObjectGlobal.QueryRow(sqlStatement, id)
 	var bookToReturn Book
-	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt)
+	err := foundRow.Scan(&bookToReturn.ID, &bookToReturn.Name, &bookToReturn.Price, &bookToReturn.Inventory, &bookToReturn.CreatedAt, &bookToReturn.UpdatedAt, &bookToReturn.Archived)
 	if err != nil {
 		switch err {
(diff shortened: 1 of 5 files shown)


cmd/api/database_test.go
```diff
@@ -272,6 +272,24 @@ func TestListBooks(t *testing.T) {
...
+			pg_tables
+		WHERE 
+			schemaname = 'public' AND 
+			tablename  = 'bookstable'
+		);`
+	check := dbObjectGlobal.QueryRow(sqlStatement)
+	var tableExists bool
+	err = check.Scan(&tableExists)
+	is.NoErr(err)
+	is.True(tableExists == false)
```
```suggestion
	is.True(!tableExists)

File: cmd/api/database_test.go
Comment (cesarFuhr) [inline cmd/api/database_test.go:1]: I am curious if we can guarantee that the migrations down tests runs in the end or if they could mess up the other tests because it ran in the middle.

**Comment by ramon-reichert** [comment]
Is there anything missing here?

**Comment by cesarFuhr** [comment]
I left the review uncommitted, sorry 😔. I did it yesterday morning.

File: cmd/api/database_test.go
Comment (ramon-reichert) [reply cmd/api/database_test.go:1]: Description of t.Run: func (*testing.T).Run(name string, f func(t *testing.T)) bool
Run runs f as a subtest of t called name. It runs f in a separate goroutine and **blocks until f returns**(MY HIGHLIGHTS) or calls t.Parallel to become a parallel test. Run reports whether f succeeded (or at least did not fail before calling t.Parallel).

Run may be called simultaneously from multiple goroutines, but all such calls must return before the outer test function for t returns.

File: cmd/api/database.go [hunk 2]
```diff
@@ -42,6 +44,7 @@ func migrationUp() error {
 	if err != nil {
 		return fmt.Errorf("migrating up: %w", err)
 	}
+	mGlobal = m
 
 	m.Up()
```
cesarFuhr: Not in this PR scope, but I think we forgot to get the error back from the function here.
```suggestion
	err = m.Up()

File: cmd/api/database_test.go
Comment (cesarFuhr) [reply cmd/api/database_test.go:1]: Ok, just make sure the Database is correctly set up before running the other tests.
Migrate the database up after the migration down tests has run.

File: cmd/api/database_test.go [hunk 3]
```diff
@@ -272,6 +272,24 @@ func TestListBooks(t *testing.T) {
 	})
 }
 
+func TestDownMigrations(t *testing.T) {
+	is := is.New(t)
```
cesarFuhr, line 352:
```suggestion
	is := is.New(t)
        t.Cleanup(func() {
                is.NoErr(mGlobal.Up())
        })

cesarFuhr (approved):

PR #23, opened 2023-09-25
Refac/layering

ramon-reichert:
The packages are separated, but the responsibilities not yet. If you can we could talk tomorrow to finish it together.


ramon-reichert:
I've written doubts on comments at main.go and handlers.go.

cmd/api/book/service.go
```diff
@@ -0,0 +1,51 @@
+package book
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type ServiceAPI interface {
+	ArchiveStatusBook(id uuid.UUID) (Book, error)
```
cesarFuhr:
```suggestion
	ArchiveBook(id uuid.UUID) (Book, error)

File: cmd/api/book/service.go [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+	CreateBook(bookEntry Book) (Book, error)
+	GetBook(id uuid.UUID) (Book, error)
+	UpdateBook(bookEntry Book, id uuid.UUID) (Book, error)
+}
+
+type Repository interface {
+	ArchiveStatusBook(id uuid.UUID, archived bool) (Book, error)
```
cesarFuhr:
```suggestion
	SetBookArchiveStatus(id uuid.UUID, archived bool) (Book, error)

File: cmd/api/http/handlers.go [hunk 3]
```diff
@@ -0,0 +1,367 @@
...
+
+type BookHandler struct {
+	bookService book.ServiceAPI
+}
+
+func NewBookHandler(bookService book.ServiceAPI) *BookHandler {
+	return &BookHandler{bookService: bookService}
+}
+
+//==========HTTP ADDRESSERS:===========
```
cesarFuhr: I don't think this name is accurate. Maybe router or mux.
I actually would start to question the need of these section comments, are they needed now that our code is spread in more files?

ramon-reichert: I can take it of.

cmd/api/book/service.go
ramon-reichert: Ok. I've got it. Changing...

ramon-reichert:
We left behind the moving of json tags to http package. I'll take a look at this.

cmd/api/book/book.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+
+	pagesTotal = int(math.Ceil(float64(itemsTotal) / float64(pageSize)))
+	if page > pagesTotal {
+		return 0, 0, 0, ErrResponseQueryPageOutOfRange
+	}
+
+	return pagesTotal, page, pageSize, nil
+}
```
cesarFuhr: This is not related to the book entity at all, I don't see any mention of a book in these pagination/total parameters functions.
Please move this to the handlers files.

Makefile
cesarFuhr, line 1: Please remove the formatting changes, they aren't adding anything.

cmd/api/book/service.go
```diff
@@ -0,0 +1,147 @@
...
+
+type PagedBooks struct {
+	PageCurrent int    `json:"page_current"`
+	PageTotal   int    `json:"page_total"`
+	PageSize    int    `json:"page_size"`
+	ItemsTotal  int    `json:"items_total"`
+	Results     []Book `json:"results"`
+}
+
+func (s *Service) ListBooks(query url.Values) (PagedBooks, error) {
```
cesarFuhr: You shouldn't be dealing with `url.Values` here. This is specific to the HTTP layer.
What you can do to pass the search parameters to the `ListBooks` method is create a `request` type.
```go
type ListBooksRequest struct {
    Name string
    MinPrice float
//...
}

Comment (cesarFuhr) [reply cmd/api/book/service.go]: Then you would extract the values from the `http.Request` in the HTTP layer.

File: cmd/api/database/database.go [hunk 6]
```diff
@@ -1,76 +1,90 @@
...
+			return book.Book{}, fmt.Errorf("archiving on db: %w", err)
 		}
 	}
 
 	return bookToReturn, nil
 }
 
-func countRows(name string, minPrice32, maxPrice32 float32, archived bool) (int, error) {
+/* Counts how many rows in db fit the specified filter parameters. */
+func (store *Store) CountRows(name string, minPrice32, maxPrice32 float32, archived bool) (int, error) {
```
cesarFuhr: This method needs to be closer to the `ListBooks` method.
I think a better name would be `CountListBooksTotals` or simply `ListBooksTotals`.

cmd/api/database/database.go
```diff
@@ -92,8 +106,42 @@ func countRows(name string, minPrice32, maxPrice32 float32, archived bool) (int,
...
+	if err != nil {
+		return book.Book{}, fmt.Errorf("storing on db: %w", err)
+	}
+
+	return bookToReturn, nil
+}
+
+/* Searches a book in database based on ID and returns it if succeed. */
+func (store *Store) GetBookByID(id uuid.UUID) (book.Book, error) {
+	sqlStatement := `SELECT id, name, price, inventory, created_at, updated_at, archived FROM bookstable WHERE id=$1;`
```
cesarFuhr: Please format this `sql` query like you did with the others. They read way better.

cmd/api/main.go
```diff
@@ -1,433 +1,57 @@
...
 	}
-
-	return sortBy, sortDirection, true
 }
 
-func main() {
+func run() error {
 	//connect to db:
-	dbObject, err := connectDb()
+	connStr := os.Getenv("DATABASE_URL") //I'M STILL CONFUSED ABOUT THAT. IN DATABASE_TEST WE CHANGED THE ENV BY A LITERAL
```
cesarFuhr: We can change the test code to read from the environment too, would be a nice improvement.

cesarFuhr (changes requested):

Makefile
ramon-reichert: Ok. I did not even see the changes. It is just blank spaces added accidentally.

cmd/api/main.go
ramon-reichert: I took a look at the old commits. now, to remember how we used to deal with the connStr in database_test before.
I've got it:
We used to SET (not GET!!) the connStr as an ENV because ConnectDB() did not receive anything:
`func TestMain(m *testing.M) {
	os.Setenv("DATABASE_URL", "postgres://root:root@localhost:5432/booksdb?sslmode=disable")
	db, err := ConnectDb()`

Now that `database.ConnectDb(connStr)` receive the connStr, we do not need the ENV anymore.

cmd/api/database/database.go
ramon-reichert: ok, done.

cmd/api/database/database.go
ramon-reichert: Done.

Makefile
ramon-reichert: Ok. I did not even see the changes. It is just blank spaces added accidentally.

cmd/api/book/service.go
ramon-reichert: I've tried to pass a variable of type http.ListBooksRequest to service.go, but then it felt in a cycle import error. So I had to pass the list of parameters extensively. It worked, at least.

cmd/api/book/book.go
ramon-reichert: Ok.

cmd/api/book/service.go
ramon-reichert: In the middle of the night I realized that I could create the struct type at service.go side!

cmd/api/main.go
cesarFuhr: Yeah, but its better to control the connection string from the outside.
We can have a quick pairing session for me to show you why is that.

Makefile
```diff
@@ -12,3 +12,6 @@ test-env-up:
 
 test-env-down:
 	docker compose -f docker-compose.test.yml down
+
+
+
```
cesarFuhr: Remove this ones too.

cmd/api/book/book.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Book struct { //IS IT POSSIBLE TO MOVE JSON TAGS TO HTTP PACKAGE??
+	ID        uuid.UUID `json:"id"`
+	Name      string    `json:"name"`
+	Price     *float32  `json:"price"`
+	Inventory *int      `json:"inventory"`
+	CreatedAt time.Time `json:"-"`
+	UpdatedAt time.Time `json:"-"`
+	Archived  bool      `json:"archived"`
+}
```
cesarFuhr, line 19: We should, but I want to finish this refactor first, we can deal with that later. I will make sure we have a ticket to address it.

cmd/api/http/handlers.go
```diff
@@ -0,0 +1,325 @@
...
+		archived = true
+	}
+
+	page, pageSize, valid := extractPageParams(query)
+	if !valid {
+		responseJSON(w, http.StatusBadRequest, book.ErrResponseQueryPageInvalid)
+		return
+	}
+
+	pagedBooks, err := h.bookService.ListBooks(name, minPrice32, maxPrice32, sortBy, sortDirection, archived, page, pageSize)
```
cesarFuhr: Define a request type as we discussed.
It should be defined inside the `book` package, that will avoid the cyclic dependency issue you had.

cesarFuhr (changes requested):
Almost there.

cmd/api/book/book.go
cesarFuhr: https://trello.com/c/scwnaydr/56-refactor-define-request-types

cmd/api/book/book.go
```diff
@@ -0,0 +1,45 @@
...
+type ListBooksRequest struct {
+	Name          string
+	MinPrice      float32
+	MaxPrice      float32
+	SortBy        string
+	SortDirection string
+	Archived      bool
+	Page          int
+	PageSize      int
+}
```
cesarFuhr: Please move this definition closer to the method that uses it.

cesarFuhr (changes requested):
One last change.

cesarFuhr (approved):

PR #24, opened 2023-10-20
shutdown 10 seconds after os.signal

ramon-reichert:
Like I left commented, we should think if 10 seconds is the right time to wait.
The code is working, but I think we should test it together with real interruptions and error simulations.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"strconv"
 	"syscall"
 	"time"
 
@@ -66,16 +65,16 @@ func run() error {
 	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM)
 	<-sc
 
-	deadline := 10
-	deadlineStr := os.Getenv("DEADLINE")
-	if deadlineStr != "" {
-		deadline, err = strconv.Atoi(deadlineStr)
+	timeout := time.Duration(10) * time.Second
+	timeoutStr := os.Getenv("SERVICE_SHUTDOWN_TIMEOUT") //This ENV must be written with a unit suffix, like seconds
+	if timeoutStr != "" {
+		timeout, err = time.ParseDuration(timeoutStr)
 		if err != nil {
-			return fmt.Errorf("getting deadline from env: %w", err)
+			return fmt.Errorf("getting shutdown timeout from env: %w", err)
 		}
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(deadline)*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout))
 	defer cancel()
 	if err := server.Shutdown(ctx); err != nil {
 		return fmt.Errorf("HTTP shutdown error: %w", err)
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -1,11 +1,15 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/books-service/cmd/api/book"
 	"github.com/books-service/cmd/api/database"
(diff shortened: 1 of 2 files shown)


cmd/api/main.go
```diff
@@ -49,9 +53,23 @@ func run() error {
...
-	err = server.ListenAndServe()
-	if err != nil && !errors.Is(err, http.ErrServerClosed) {
-		return fmt.Errorf("unexpected http server error: %w", err)
+	go func() (err error) {
+		err = server.ListenAndServe()
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			return fmt.Errorf("unexpected http server error: %w", err)
+		}
+		return nil
+	}()
```
cesarFuhr: I don't think this function should be returning an error. You never read it.
```suggestion
	go func() {
		err = server.ListenAndServe()
		if err != nil && !errors.Is(err, http.ErrServerClosed) {
			log.Println("unexpected http server error: %v", err)
		}
		log.Println("stopped serving new requests.")
	}()

File: cmd/api/main.go [hunk 2]
```diff
@@ -49,9 +53,23 @@ func run() error {
...
+
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM)
+	<-sc
+
+	ctx, shutdownRelease := context.WithTimeout(context.Background(), 10*time.Second) //HOW MUCH TIME MAKE SENSE FOR US??
+	defer shutdownRelease()
```
cesarFuhr: I don't like the cancel function name here (`shutdownRelease`)... I would stick with the usual `cancel` or go for something like `shutdownCancel`.

cesarFuhr (changes requested):

cesarFuhr:
Nice job, I would like to  talk to you about this.
1 - How do you tested it?
2 - Do you know how this actually works?
3 - Have you done research around `channels`and `goroutines`?
4 - Does this code has concurrent execution?

cmd/api/main.go
```diff
@@ -49,9 +53,23 @@ func run() error {
...
+			return fmt.Errorf("unexpected http server error: %w", err)
+		}
+		return nil
+	}()
+
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM)
+	<-sc
+
+	ctx, shutdownRelease := context.WithTimeout(context.Background(), 10*time.Second) //HOW MUCH TIME MAKE SENSE FOR US??
```
cesarFuhr: This timeout value should be configurable and we should read it from an environment variable. We also should have an default value, that could be 10 seconds.

ramon-reichert: Ok. We can talk this afternoon if you want. Just tell me at whatsapp.

cmd/api/main.go
```diff
@@ -49,9 +54,32 @@ func run() error {
...
+		}
+		log.Println("stopped serving new requests.")
+	}()
+
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, syscall.SIGINT, syscall.SIGTERM)
+	<-sc
+
+	deadline := 10
+	deadlineStr := os.Getenv("DEADLINE")
```
cesarFuhr: Lets go with `SERVICE_SHUTDOWN_TIMEOUT`.

cesarFuhr: Lets use the `time.Duration` type. You can use this function to parse it https://pkg.go.dev/time#ParseDuration .

cmd/api/main.go
```diff
@@ -49,9 +54,32 @@ func run() error {
...
+	if deadlineStr != "" {
+		deadline, err = strconv.Atoi(deadlineStr)
+		if err != nil {
+			return fmt.Errorf("getting deadline from env: %w", err)
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(deadline)*time.Second)
```
cesarFuhr: If you use duration you wont need to multiply it by seconds here.

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
     environment:
       DATABASE_URL: "postgres://postgres:chevas@db:5432/booksdb?sslmode=disable"
       DATABASE_MIGRATIONS_PATH: "/src/migrations"
+      DEADLINE: "10"
```
cesarFuhr: Don't forget to change it here.
You can see why I am asking you to change the env variable name by looking at this env declaration in the docker-compose file here. What deadline stands for? You cannot tell looking from the outside. You would have to inspect the code to make sure what it actually represents.

cesarFuhr (changes requested):

cesarFuhr (approved):

PR #25, opened 2023-11-07
Refac/request types

ramon-reichert:


cmd/api/book/service.go
```diff
@@ -37,23 +37,50 @@ func (s *Service) ArchiveBook(id uuid.UUID) (Book, error) {
 	return s.repo.SetBookArchiveStatus(id, archived)
 }
 
-func (s *Service) CreateBook(bookEntry Book) (Book, error) {
-	bookEntry.ID = uuid.New()                                      //Atribute an ID to the entry
-	bookEntry.CreatedAt = time.Now().UTC().Round(time.Millisecond) //Atribute creating and updating time to the new entry. UpdateAt can change later.
-	bookEntry.UpdatedAt = bookEntry.CreatedAt
-	return s.repo.CreateBook(bookEntry)
+type EntryBookRequest struct {
```
cesarFuhr:
```suggestion
type CreateBookRequest struct {

File: cmd/api/book/service.go [hunk 2]
```diff
@@ -37,23 +37,50 @@ func (s *Service) ArchiveBook(id uuid.UUID) (Book, error) {
...
+	Name      string
+	Price     *float32
+	Inventory *int
+}
+
+func (s *Service) CreateBook(bookEntry EntryBookRequest) (Book, error) {
```
cesarFuhr:
```suggestion
func (s *Service) CreateBook(req CreateBookRequest) (Book, error) {

File: cmd/api/book/service.go [hunk 3]
```diff
@@ -37,23 +37,50 @@ func (s *Service) ArchiveBook(id uuid.UUID) (Book, error) {
...
+	createdAt := time.Now().UTC().Round(time.Millisecond) //Atribute creating and updating time to the new entry. UpdateAt can change later.
+	newBook := Book{
+		ID:        uuid.New(), //Atribute an ID to the entry
+		Name:      bookEntry.Name,
+		Price:     bookEntry.Price,
+		Inventory: bookEntry.Inventory,
+		CreatedAt: createdAt,
+		UpdatedAt: createdAt,
+		//Archived is set to false by defalut inside database
```
cesarFuhr, line 55: Achive is `false` by default since the zero value of a boolean is false.

cmd/api/book/service.go
```diff
@@ -37,23 +37,50 @@ func (s *Service) ArchiveBook(id uuid.UUID) (Book, error) {
...
+	}
+	return s.repo.CreateBook(newBook)
+}
+
+func (s *Service) UpdateBook(bookEntry EntryBookRequest, id uuid.UUID) (Book, error) {
```
cesarFuhr: This is weird, there is no reason to pass the `uuid` separately, this looks like you trying to use the same type for the create and update methods. Just write a new type for the UpdateBook request.
```suggestion
type UpdateBookRequest struct {
        ID        uuid.UUID
	Name      string
	Price     *float32
	Inventory *int
}

func (s *Service) UpdateBook(req UpdateBookRequest) (Book, error) {
Comment

File: cmd/api/http/handlers.go [hunk 5]
```diff
@@ -256,6 +256,49 @@ func isolateId(w http.ResponseWriter, r *http.Request) (id uuid.UUID, err error)
...
+	ItemsTotal  int            `json:"items_total"`
+	Results     []BookResponse `json:"results"`
+}
+
+/*Copy the fields of a PagedBooks object to an http layer struct with json tags*/
+func pagedBooksToResponse(page book.PagedBooks) PageOfBooksResponse {
+	jsonTagedBooks := []BookResponse{}
+	for _, book := range page.Results {
+		jsonTagedBooks = append(jsonTagedBooks, bookToResponse(book))
+	}
```
cesarFuhr: I am not sure about the variable name here. `jsonTagedBooks` looks like a little too much, maybe use `results` or `bookResults`. This is a pretty short lived function, you don't need to have such big names.

cmd/api/http/handlers.go
```diff
@@ -75,12 +75,12 @@ func (h *BookHandler) archiveBook(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	responseJSON(w, http.StatusOK, archivedBook)
+	responseJSON(w, http.StatusOK, bookToResponse(archivedBook))
 }
 
 /* Validates the entry, then stores the entry as a new book. */
 func (h *BookHandler) createBook(w http.ResponseWriter, r *http.Request) {
-	var bookEntry book.Book
+	var bookEntry book.EntryBookRequest
```
cesarFuhr: Have you tested this?
`book.EntryBookRequest` does not have the `json` tags, and I don't think it should. You can create new request types for the `http` layer here.

cesarFuhr (changes requested):

cmd/api/http/handlers.go
ramon-reichert: Yes, I've tested with Postman yesterday and now again. Why should this fail?

cmd/api/book/service.go
ramon-reichert: Yes, I was exactly trying to use the same type for both methods. Thas why I've called the type EntryBookRequest. But I will separate them. No problem.

cmd/api/book/service.go
ramon-reichert: Yes, but is set at the migrations, too: 
```
ALTER TABLE public.bookstable
  ADD COLUMN IF NOT EXISTS archived BOOLEAN DEFAULT false;
```
I've put this comment just to keep in mind that this field is present too, although it's not being initialized in this function. Do you think should I take it off?

cesarFuhr: well... there is no harm in the comment, just notice that if you add such a comment you may start to rely on them.
Then when you forget to add it for some reason and it will make your life harder.

cmd/api/http/handlers.go
ramon-reichert: Thats because of the json Decoder? It accepts case-insensitive match, thats why it is working, right?

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match. By default, object keys which don't have a corresponding struct field are ignored (see Decoder.DisallowUnknownFields for an alternative).

But, if it's better, I can create an http layer request type exclusively for the decoding, and then copy its fields to the book layer struct, trouthg a function just like `bookToResponse`, but the opposite. You say...

cesarFuhr: Its probably OK because of the go std lib default behavior, which is converting the field names to camel case when decoding a `json` object into a struct. Sometimes that will not be enough. I would probably create the separate struct to be more consistent.

cmd/api/book/service.go
ramon-reichert: I thought is was no problem, but I'm facing some of them, since we have different structs to Update and Create. The function `FilledFields`, that is defined at `service.go`, now has to deal with this different types, and I'm having to write a lot of code just to ajust the incoming book data to pass it to this little function. It makes more sense to me if we transfer this function (or its functionality) to the http layer. It could be a function that verifies the fields and already copy them to the proper struct (CreateBookResquest or UpdateBookRequest).
Ooooor we can come back to having just one type for both methods.

cesarFuhr: You should have separate functions to validate each one of the structs, also you can move it to the handlers layer if you want to.

ramon-reichert:
Just waiting your approval... =)

cesarFuhr (approved):

PR #26, opened 2023-11-21
Feat/service tests

ramon-reichert:
The service test that we started togheter is working, but the database test is failling.

$ go test -p=1 ./...
?       github.com/books-service/cmd/api        [no test files]
ok      github.com/books-service/cmd/api/book   (cached)
?       github.com/books-service/cmd/api/book/mocks     [no test files]
2023/11/20 22:06:48 connecting to db, pingging: pq: SSL is not enabled on the server
FAIL    github.com/books-service/cmd/api/database       0.277s
?       github.com/books-service/cmd/api/http   [no test files]

 Even switching to older commits, the error persists, so it seems not to be fault of our new code. I will move forward with the service tests, but sometime we will have to solve this problem with the db connection. Some idea?


ramon-reichert:
That's weird. I've added some debug code before calling func `ConnectDb` and found out here that bash was not setting the ENV variables right, so I've tried to set them with `export` command, and it worked. But here this commmand was not welcome.

cesarFuhr:
I am curious on how you ran the make target in your machine. I reverted the `export` changes and it worked as expected.
Also I did a little improvement on the time comparison test, turns out we are rounding to milliseconds and that creates some weird edge cases when using after/before.
Check the `Compare` method documentation to understand it better.
https://pkg.go.dev/time#Time.Compare

ramon-reichert:
> I am curious on how you ran the make target in your machine...

Makefile didn't work in my machine, do you remember? I'm copying each line and pasting it at terminal.

cesarFuhr:
> > I am curious on how you ran the make target in your machine...
> 
> Makefile didn't work in my machine, do you remember? I'm copying each line and pasting it at terminal.

You need to run all the lines at once, otherwise the environment variables wont load before the `go test` command.

ramon-reichert:
> You need to run all the lines at once, otherwise the environment variables wont load before the `go test` command.

Nice. It worked. But, why? Isn't the environment the same?

ramon-reichert:
Missing some edge tests for ListBooks method.
- itemsTotal == 0 case;
- Any other you want to add?

cesarFuhr:
Nice job, as an exercise I want you to write a test where the database returns an error.
Maybe in the ListBooks method (which is one of the more complex cases).

There is no need to write a test for every method, one will be enough since it is an exercise.

cesarFuhr:
Missing some edge tests for ListBooks method.

    itemsTotal == 0 case;
    Any other you want to add?

cesarFuhr (approved):

PR #27, opened 2023-11-25
Feat/http tests

ramon-reichert:


ramon-reichert:
@cesarFuhr  Any comments until here?

cesarFuhr:
This looks good, I would ask you to change this in two ways:
1 - Use directly the handle functions, it will simplify the code without loosing coverage.
2 - Assert that the JSON response (in the routes that do respond something) are being correctly written to the response body.

ramon-reichert:
> 1 - Use directly the handle functions, it will simplify the code without loosing coverage. 

To directly call the handle function I had to export it. Isn't that right, right??
Furthermore, this way we do not test the http route itself, like I left at comments...

I think I'm loosing something...

ramon-reichert:
> 2 - Assert that the JSON response (in the routes that do respond something) are being correctly written to the response body.

I'll move on to implement this change, while waiting your help on the previous topic.

cesarFuhr:
Ok, its not unusual to have the handler functions in a separate package, then you would have to export them anyway.
You may have a single server to several APIs that use different set of handlers, that's when you need to test them separately.

You know what, keep it as it was, with the tests calling the server. Where did you got that idea? (its a good one :+1: )

cesarFuhr:
Please try to clean up the PR, all the changes from the previous one are being listed here.
Why didn't you branched off `main` when you started this piece of work?

ramon-reichert:
> Ok, its not unusual to have the handler functions in a separate package, then you would have to export them anyway. You may have a single server to several APIs that use different set of handlers, that's when you need to test them separately.
> 
> You know what, keep it as it was, with the tests calling the server. Where did you got that idea? (its a good one 👍 )

It's not my idea. Thats what I've understood from the reference you gave me: 
https://quii.gitbook.io/learn-go-with-tests/build-an-application/http-server

I'll keep this way, so.

ramon-reichert:
> Please try to clean up the PR, all the changes from the previous one are being listed here. Why didn't you branched off `main` when you started this piece of work?

Sorry, I forgot. It's because I've started this ticket while the previous one was not merged yet. And after the merged, I forgot to rebase.

ramon-reichert:
I've tried to clean up the commits, but its a mess. I'll open another PR.

ramon-reichert:
I've tried to clean up the commits, but its a mess. I'll open another PR.

PR #28, opened 2023-12-11
Feat/http test

ramon-reichert:
Openning a clean PR.


ramon-reichert:
It was supposed to be clean, at least.

ramon-reichert:
> 2 - Assert that the JSON response (in the routes that do respond something) are being correctly written to the response body.

I'll continue to work on it.

ramon-reichert:
Please, help me if you can. I've put this debug lines trying to find the problem while comparing the JSON response body with what  it should be.
The thing is: In the body response, there is an extra "new line" at the end. Why is that?

ramon-reichert:
I think I found:
At func `responseJSON` we are using `json.Encode()`, whose description is: 

> func` (*json.Encoder).Encode(v any) error
Encode writes the JSON encoding of v to the stream, followed by a newline character.

Then, the question is: This newline character is ok? Because if so, I can add it to the comparision in the tests and everything will be fine...

cmd/api/http/http_test.go
```diff
@@ -0,0 +1,74 @@
...
+			ID:        newID,
+			Name:      reqBook.Name,
+			Price:     reqBook.Price,
+			Inventory: reqBook.Inventory,
+			CreatedAt: time.Now().UTC().Round(time.Millisecond),
+			UpdatedAt: time.Now().UTC().Round(time.Millisecond),
+			Archived:  false,
+		}
+		expectedJSONresponse := fmt.Sprintf(`{"id":"%s","name":"HTTP tester book","price":100,"inventory":99,"archived":false}
+`, newID)
```
cesarFuhr:
```suggestion
		expectedJSONresponse := fmt.Sprintf(`{"id":"%s","name":"HTTP tester book","price":100,"inventory":99,"archived":false}\n`, newID)

**Comment by cesarFuhr** [comment]
This looks good, please in the next use `json.Marshal` to encode the expected response. I know it seems weird, but try it and tel me what your thoughts will be.
This is all nice and looking good. Select a few more handlers to test, there is no need to do all of them. I think you already got the essence of it.

File: cmd/api/http/http_test.go (hunk 1 above)
Comment (ramon-reichert) [reply cmd/api/http/http_test.go]: It did not work. Sprintf do not interprets `\n`

Comment (cesarFuhr) [reply cmd/api/http/http_test.go]: Its not `Sprintf` its the back ticks... ok. Leave it as it was.

Comment (ramon-reichert) [reply cmd/api/http/http_test.go]: Yes, thats what I meant: `Sprintf` doesnt interpret the `\n` INSIDE the back ticks.

**Comment by ramon-reichert** [comment]
I created tests to ListBooks method, that is the harder one.
I've used json.Marshall, as you asked me, and found the same error: It not put the extra line as the json.Encode. So it was solved with this append: `expectedJSONresponse = append(expectedJSONresponse, 10)`

**Comment by ramon-reichert** [comment]
I think its done, unless if you want me to write tests to any method from the other route (`/book`), which redirects to the methods `getBookById`, `updateBook` and `archiveBook`.

File: cmd/api/http/http_test.go (hunk 1 above)
Comment (cesarFuhr) [reply cmd/api/http/http_test.go]: Actually, the back ticks escape the `\`, so it is interpreted as if it was `\\n`, that is why `Sprintf` doesn't picks it up. But that is just standard string (back ticks) behavior.

File: cmd/api/http/http_test.go [hunk 2]
```diff
@@ -0,0 +1,308 @@
...
+			PageCurrent: params.Page,
+			PageTotal:   3,
+			PageSize:    params.PageSize,
+			ItemsTotal:  listSize,
+			Results:     testBookslist[0:10],
+		}
+
+		expectedJSONresponse, err := json.Marshal(pagedBooksToResponse(expectedReturn))
+		is.NoErr(err)
+		expectedJSONresponse = append(expectedJSONresponse, 10)
```
cesarFuhr: This is a bit obscure. What does `10` stands for? lets change it to this so we know what we are appending.
```suggestion
		expectedJSONresponse = append(expectedJSONresponse, []byte("\n")...)

File: cmd/api/http/http_test.go [hunk 3]
```diff
@@ -0,0 +1,308 @@
...
+			PageCurrent: params.Page,
+			PageTotal:   2,
+			PageSize:    params.PageSize,
+			ItemsTotal:  listSize,
+			Results:     testBookslist[15:30],
+		}
+
+		expectedJSONresponse, err := json.Marshal(pagedBooksToResponse(expectedReturn))
+		is.NoErr(err)
+		expectedJSONresponse = append(expectedJSONresponse, 10)
```
cesarFuhr:
```suggestion
		expectedJSONresponse = append(expectedJSONresponse, []byte("\n")...)

File: cmd/api/http/http_test.go [hunk 4]
```diff
@@ -0,0 +1,308 @@
...
+	})
+
+	t.Run("expected order parameters error", func(t *testing.T) {
+		is := is.New(t)
+
+		url := "/books?sort_by=invalid string"
+
+		expectedJSONresponse, err := json.Marshal(book.ErrResponseQuerySortByInvalid)
+		is.NoErr(err)
+		expectedJSONresponse = append(expectedJSONresponse, 10)
```
cesarFuhr:
```suggestion
		expectedJSONresponse = append(expectedJSONresponse, []byte("\n")...)

File: cmd/api/http/http_test.go [hunk 5]
```diff
@@ -0,0 +1,308 @@
...
+	})
+
+	t.Run("expected page parameters error", func(t *testing.T) {
+		is := is.New(t)
+
+		url := "/books?page_size=35"
+
+		expectedJSONresponse, err := json.Marshal(book.ErrResponseQueryPageInvalid)
+		is.NoErr(err)
+		expectedJSONresponse = append(expectedJSONresponse, 10)
```
cesarFuhr:
```suggestion
		expectedJSONresponse = append(expectedJSONresponse, []byte("\n")...)

cesarFuhr (approved):
Only those minor tweaks.
Nice job.

cesarFuhr:
Is it ready for review?

PR #29, opened 2023-12-28
wip context added to createBook handler

ramon-reichert:
Just beginning... Commiting so you can watch my steps as I go.
By now, I just added the ctx wiht timeout to the createBook handler and its associated functions, and change what was needed to solve problems and pass the tests.

### DIFF
--- a/cmd/api/book/book_test.go
+++ b/cmd/api/book/book_test.go
@@ -1,6 +1,7 @@
 package book_test
 
 import (
+	"context"
 	"errors"
 	"testing"
 	"time"
@@ -11,6 +12,8 @@ import (
 	"github.com/matryer/is"
 	gomock "go.uber.org/mock/gomock"
 )
+
+var ctx context.Context = context.TODO() //Should we have a specific context to run the tests, here too? This one lives forever.
 
 func TestCreateBook(t *testing.T) {
 	t.Run("creates a book without errors", func(t *testing.T) {
@@ -25,7 +28,7 @@ func TestCreateBook(t *testing.T) {
 			Inventory: toPointer(99),
 		}
 
-		mockRepo.EXPECT().CreateBook(gomock.Any()).DoAndReturn(func(b book.Book) (book.Book, error) {
+		mockRepo.EXPECT().CreateBook(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, b book.Book) (book.Book, error) {
 			is.True(b.ID != uuid.Nil)
 			is.Equal(b.Name, reqBook.Name)
 			is.Equal(b.Price, reqBook.Price)
@@ -36,7 +39,7 @@ func TestCreateBook(t *testing.T) {
 			return b, nil
 		})
 
-		createdBook, err := mS.CreateBook(reqBook)
+		createdBook, err := mS.CreateBook(ctx, reqBook)
 		is.NoErr(err)
 		is.True(createdBook.ID != uuid.Nil)
 		is.Equal(createdBook.Name, reqBook.Name)
--- a/cmd/api/book/book_test.go
+++ b/cmd/api/book/book_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"log"
-	"os"
 	"testing"
 	"time"
 
@@ -16,20 +15,6 @@ import (
 )
 
 var ctx context.Context = context.Background()
-
-func TestMain(m *testing.M) {
-	var err error
-
-	reqTimeoutStr := os.Getenv("HTTP_REQUEST_TIMEOUT") //This ENV must be written with a unit suffix, like seconds
-	if reqTimeoutStr != "" {
-		book.RequestTimeout, err = time.ParseDuration(reqTimeoutStr)
-		if err != nil {
-			log.Fatalln("getting request timeout from env: %w", err)
-		}
-	}
-
-	os.Exit(m.Run())
-}
 
 func TestCreateBook(t *testing.T) {
 	t.Run("creates a book without errors", func(t *testing.T) {
@@ -323,20 +308,21 @@ func TestListBooks(t *testing.T) {
 		}
 		itemsTotal := 30
 		results := []book.Book{}
-		ctxTest, cancel := context.WithTimeout(context.Background(), book.RequestTimeout)
+		reqTimeout := time.Duration(1) * time.Second
+		ctxTest, cancel := context.WithTimeout(context.Background(), reqTimeout)
 		defer cancel()
 
 		mockRepo.EXPECT().ListBooksTotals(gomock.Any(), reqBooks.Name, reqBooks.MinPrice, reqBooks.MaxPrice, reqBooks.Archived).Return(itemsTotal, nil)
 
 		mockRepo.EXPECT().ListBooks(gomock.Any(), reqBooks.Name, reqBooks.MinPrice, reqBooks.MaxPrice, reqBooks.SortBy, reqBooks.SortDirection, reqBooks.Archived, reqBooks.Page, reqBooks.PageSize).DoAndReturn(func(ctx context.Context, name string, minPrice32, maxPrice32 float32, sortBy, sortDirection string, archived bool, page, pageSize int) ([]book.Book, error) {
-			time.Sleep(book.RequestTimeout + time.Second)
+			time.Sleep(reqTimeout + time.Second)
 			log.Println("context error: ", ctxTest.Err())
 			return results, ctxTest.Err()
 		})
 
 		pageOfBooksList, err := mS.ListBooks(ctx, reqBooks)
 		is.Equal(pageOfBooksList, book.PagedBooks{})
-		is.Equal(err, context.DeadlineExceeded)
+		is.Equal(err.Error(), "timeout on call to ListBooks: "+context.DeadlineExceeded.Error())
 	})
 }
 
--- a/cmd/api/book/book_test.go
+++ b/cmd/api/book/book_test.go
@@ -13,7 +13,7 @@ import (
 	gomock "go.uber.org/mock/gomock"
 )
 
-var ctx context.Context = context.TODO() //Should we have a specific context to run the tests, here too? This one lives forever.
+var ctx context.Context = context.Background()
 
 func TestCreateBook(t *testing.T) {
 	t.Run("creates a book without errors", func(t *testing.T) {
--- a/cmd/api/book/book_test.go
+++ b/cmd/api/book/book_test.go
@@ -3,7 +3,6 @@ package book_test
 import (
 	"context"
 	"errors"
-	"log"
 	"testing"
 	"time"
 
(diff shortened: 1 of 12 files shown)


ramon-reichert:
Still have to implement the error check, define better the context used at the tests, and make this to all the handlers and methods.
- Can y use a global and exported timeout constant to all packages, defined at service.go?
- The link of that article didnt come by whatsapp...

cmd/api/http/handlers.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/google/uuid"
 )
 
-var timeout = time.Duration(1) * time.Millisecond
+var timeout = time.Duration(30) * time.Millisecond
```
ramon-reichert: Set 30 mS to test timeout with Postman. I think we could set it with 5 seconds to normal operation.

cmd/api/http/handlers.go
```diff
@@ -122,6 +122,14 @@ func (h *BookHandler) createBook(w http.ResponseWriter, r *http.Request) {
 
 	storedBook, err := h.bookService.CreateBook(r.Context(), reqBook)
 	if err != nil {
+		if errors.Is(err, r.Context().Err()) {
```
ramon-reichert: This way we can catch a timed out context as well a canceled one.

cmd/api/http/handlers.go
```diff
@@ -122,6 +122,14 @@ func (h *BookHandler) createBook(w http.ResponseWriter, r *http.Request) {
...
+			errCtx := book.ErrResponse{
+				Code:    book.ErrResponseFromContext.Code,
+				Message: book.ErrResponseFromContext.Message + r.Context().Err().Error(),
+			}
+			responseJSON(w, http.StatusInternalServerError, errCtx)
```
ramon-reichert: I' even thought about using a `408 StatusRequestTimeout` , but it denotes an error from client, not from server:

> [15.5.9. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.9)[408 Request Timeout](https://www.rfc-editor.org/rfc/rfc9110.html#name-408-request-timeout)
The 408 (Request Timeout) status code indicates that the server did not receive a complete request message within the time that it was prepared to wait.
If the client has an outstanding request in transit, it MAY repeat that request. If the current connection is not usable (e.g., as it would be in HTTP/1.1 because request delimitation is lost), a new connection will be used.

So, I think a 500 fits better here.

cesarFuhr: 504 is the best here. It's a gateway timeout, but it's widely used as a server timeout too.

ramon-reichert: ok

ramon-reichert:
> Still have to implement the error check, define better the context used at the tests, and make this to all the handlers and methods.
> 
> * Can y use a global and exported timeout constant to all packages, defined at service.go?
> * The link of that article didnt come by whatsapp...

And what about this things?

cmd/api/book/book_test.go
```diff
@@ -12,6 +13,8 @@ import (
 	gomock "go.uber.org/mock/gomock"
 )
 
+var ctx context.Context = context.TODO() //Should we have a specific context to run the tests, here too? This one lives forever.
```
cesarFuhr: Only if you are testing cancellation through context.

cmd/api/database/database_test.go
```diff
@@ -23,6 +24,7 @@ import (
 
 var store *database.Store
 var sqlDB *sql.DB
+var ctx context.Context = context.TODO() //Should we have a specific context to run the tests?
```
cesarFuhr: Only if you are testing cancellation through context.

cmd/api/http/handlers.go
cesarFuhr: This should be read from the environment as configuration.

cmd/api/http/handlers.go
cesarFuhr: You can have two types of Context error => https://pkg.go.dev/context#pkg-constants.
When context is cancelled the request was `Aborted`.
When context deadline was exceeded its a `Timeout`.

`context.Cancelled`
`context.DeadlineExceeded`

cmd/api/http/http_test.go
```diff
@@ -18,6 +19,8 @@ import (
 	"go.uber.org/mock/gomock"
 )
 
+var ctx context.Context = context.TODO() //Should we have a specific context to run the tests?
```
cesarFuhr: Its OK to use context background in tests.

cmd/api/book/errors.go
```diff
@@ -18,3 +18,4 @@ var ErrResponseQuerySortByInvalid = ErrResponse{105, "query parameter 'sort_by'
 var ErrResponseQueryPageInvalid = ErrResponse{106, "query parameter 'page' must be an int starting in 1. 'page_size' must be an int beetween 1 and 30."}
 var ErrResponseQueryPageOutOfRange = ErrResponse{107, "page out of range."}
 var ErrResponseFromRespository = ErrResponse{108, "error from repository call:"}
+var ErrResponseFromContext = ErrResponse{109, "error from context:"}
```
cesarFuhr: I am not sure about this.
Maybe create two errors:
`ErrResponseRequestTimeout`
`ErrResponseRequestAborted`

cmd/api/http/http_test.go
ramon-reichert: > You can have two types of Context error => https://pkg.go.dev/context#pkg-constants.
When context is cancelled the request was Aborted.
When context deadline was exceeded its a Timeout.
context.Cancelled
context.DeadlineExceeded

Yes, I saw this. Thas why I created just one ErrResponse to both cases. `r.Context().Err()` will return the especific error (cancelled or deadline) occured and it will be showed at http response, added to `ErrResponseFromContext`. It simplifies the code, I see.

cmd/api/http/handlers.go
ramon-reichert: I put both error types at `http_test.go` at the new commit to we see how it works.

ramon-reichert:
I tested setting env HTTP_REQUEST_TIMEOUT to 30ms to see the context timeout error response at Postman. It worked.

ramon-reichert:
Now I'm waiting that article you would send to me, then I will extend context handling to all methods.

cmd/api/book/service.go
```diff
@@ -1,27 +1,31 @@
...
 import (
+	"context"
+	"errors"
 	"math"
 	"time"
 
 	"github.com/google/uuid"
 )
 
+var RequestTimeout = time.Duration(5) * time.Second
```
cesarFuhr: This should not be a global variable. You need to add this to the `http.BookHandler` struct, its part of the service configuration.

cmd/api/book/book_test.go
```diff
@@ -12,6 +15,22 @@ import (
...
+func TestMain(m *testing.M) {
+	var err error
+
+	reqTimeoutStr := os.Getenv("HTTP_REQUEST_TIMEOUT") //This ENV must be written with a unit suffix, like seconds
+	if reqTimeoutStr != "" {
+		book.RequestTimeout, err = time.ParseDuration(reqTimeoutStr)
+		if err != nil {
+			log.Fatalln("getting request timeout from env: %w", err)
+		}
+	}
```
cesarFuhr: For testing proposes you don't need to read from the environment, you can set it directly so you can control it.

cmd/api/book/service.go
cesarFuhr: Remember, dependency injection.

cmd/api/book/service.go
```diff
@@ -101,9 +105,12 @@ type ListBooksRequest struct {
 	PageSize      int
 }
 
-func (s *Service) ListBooks(params ListBooksRequest) (PagedBooks, error) {
-	itemsTotal, err := s.repo.ListBooksTotals(params.Name, params.MinPrice, params.MaxPrice, params.Archived)
+func (s *Service) ListBooks(ctx context.Context, params ListBooksRequest) (PagedBooks, error) {
+	itemsTotal, err := s.repo.ListBooksTotals(ctx, params.Name, params.MinPrice, params.MaxPrice, params.Archived)
 	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return PagedBooks{}, err
```
cesarFuhr: Please wrap the error.

cmd/api/book/service.go
```diff
@@ -127,8 +134,11 @@ func (s *Service) ListBooks(params ListBooksRequest) (PagedBooks, error) {
 	}
 
 	//Ask filtered list to db:
-	returnedBooks, err := s.repo.ListBooks(params.Name, params.MinPrice, params.MaxPrice, params.SortBy, params.SortDirection, params.Archived, params.Page, params.PageSize)
+	returnedBooks, err := s.repo.ListBooks(ctx, params.Name, params.MinPrice, params.MaxPrice, params.SortBy, params.SortDirection, params.Archived, params.Page, params.PageSize)
 	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return PagedBooks{}, err
```
cesarFuhr: Please wrap the error.

cmd/api/http/http_test.go
```diff
@@ -18,13 +21,28 @@ import (
...
+func TestMain(m *testing.M) {
+	var err error
+
+	reqTimeoutStr := os.Getenv("HTTP_REQUEST_TIMEOUT") //This ENV must be written with a unit suffix, like seconds
+	if reqTimeoutStr != "" {
+		book.RequestTimeout, err = time.ParseDuration(reqTimeoutStr)
+		if err != nil {
+			log.Fatalln("getting request timeout from env: %w", err)
+		}
+	}
```
cesarFuhr: Same thing about the timeout not being a global and actually being part of the `BookHandler`.

cmd/api/main.go
```diff
@@ -53,6 +53,14 @@ func run() error {
 	//create and init http server:
 	server := bookhttp.NewServer(bookhttp.ServerConfig{Port: 8080}, bookHandler)
 
+	reqTimeoutStr := os.Getenv("HTTP_REQUEST_TIMEOUT") //This ENV must be written with a unit suffix, like seconds
+	if reqTimeoutStr != "" {
+		book.RequestTimeout, err = time.ParseDuration(reqTimeoutStr)
```
cesarFuhr: Again, move this to a configuration inside the `BookHandler`.

cmd/api/http/http_test.go
ramon-reichert: Ok, done. But if we want change the timeout value to watch the behavior of tests, wouldn't it be nice to get that value trought environment?

cmd/api/book/service.go
ramon-reichert: Which new information could I add here?
It because the error arrives already wrapped at this point: 

>...
> row := store.db.QueryRowContext(ctx, sqlStatement, name, minPrice32, maxPrice32, archived)
> 	var count int
> 	err := row.Scan(&count)
> 	if err != nil {
> 		return count, fmt.Errorf("counting books from db: %w", err)
> 	}
> ...

ramon-reichert: Watch which error it returned when I was testing the error route: (timeout set to 5mS to force ctx to timeout):

image

ramon-reichert: Just to talk on next call...

ramon-reichert: And we need to talk about the log.Prints too. Just to remember...

cmd/api/http/http_test.go
cesarFuhr: Your tests should receive the context for any given method call, you can setup the context to be a `WithTimeout` with the timeout you would want.

cmd/api/book/service.go
cesarFuhr: This is not correct, yeah... we need to talk about it.

cesarFuhr: > Which new information could I add here? It because the error arrives already wrapped at this point:
> 
> > ...
> > row := store.db.QueryRowContext(ctx, sqlStatement, name, minPrice32, maxPrice32, archived)
> > var count int
> > err := row.Scan(&count)
> > if err != nil {
> > return count, fmt.Errorf("counting books from db: %w", err)
> > }
> > ...

Its already wrapped in the database layer.
You should add some context in the service layer. I now it may sound repetitive, but it gives you a nice trace on where the error passed.

cmd/api/http/handlers.go
```diff
@@ -421,3 +423,18 @@ func extractPageParams(query url.Values) (page, pageSize int, valid bool) {
...
+		return
+	}
+	if errors.Is(err, book.ErrResponseBookNotFound) {
+		log.Println(err)
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	log.Println(err)
+	w.WriteHeader(http.StatusInternalServerError)
+}
```
cesarFuhr: You could use this function to avoid having to write this error handling in every http handler.

cesarFuhr (approved):
One nit pick only.

cmd/api/http/handlers.go
ramon-reichert: The only method that was not using `handleError` was listBooks because of the default (unknow) case of it was a BadRequest. Now that we change it to an InternalServerError, like in the other methods, it fits `handleError` too.

PR #31, opened 2024-01-11
Feat/send notifications

ramon-reichert:
Waiting your review to continue. I'll keep working on tests...


ramon-reichert:
Still missing tests to service layer mocking the notifier.

cmd/api/book/book_test.go
```diff
@@ -54,13 +43,22 @@ func TestCreateBook(t *testing.T) {
 			return b, nil
 		})
 
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		mockNtfy.EXPECT().BookCreated(gomock.Any(), reqBook.Name, *reqBook.Inventory).DoAndReturn(func(ctx context.Context, title string, inventory int) error {
+			defer wg.Done()
+			return errors.New("error to test ntfy return")
```
ramon-reichert: This error will not be actually tested, because after return, it is just logged with `log.Println` . I even tried to capture de log at stdout, but then I thought that I was going to far to this simple test. Don't you think?

cmd/api/notifications/notifications.go
```diff
@@ -6,37 +6,33 @@ import (
...
 		}
 		_, err = ntf.client.Do(req)
 		if err != nil {
-			return fmt.Errorf("error delivering message ("+fmt.Sprintf("New book created:\nTitle: %s\nInventory: %v", title, inventory)+") to topic ("+ntf.baseURL+"/New_book_created): %w", err)
+			return fmt.Errorf("error delivering message ("+fmt.Sprintf("New book created: Title: %s Inventory: %v", title, inventory)+"): %w", err)
 		}
+		return nil
 	}
 	return errors.New("notifications not enabled")
 }
```
ramon-reichert, line 54: With code the way it is, after we send the notification, we are not listening to the topic to certify that the message was delivered. We could do this with a GET request, like I did in notifications_test.go. Do you think its necessary?

cmd/api/book/book_test.go
cesarFuhr: Yeah, its good enough. There is no need to test the standard library

cmd/api/book/book_test.go
```diff
@@ -14,13 +15,16 @@ import (
 )
 
 var ctx context.Context = context.Background()
+var notificationsTimeout = 2 * time.Second
```
cesarFuhr: This can be a `const`.

cmd/api/book/book_test.go
cesarFuhr: Why are you returning an error though? Aren't you simulating a success?

cmd/api/book/service.go
```diff
@@ -57,7 +68,17 @@ func (s *Service) CreateBook(ctx context.Context, req CreateBookRequest) (Book,
...
 	}
-	return s.repo.CreateBook(ctx, newBook)
+
+	b, err := s.repo.CreateBook(ctx, newBook)
+	if err == nil {
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), s.notificationsTimeout)
+			defer cancel()
+			err := s.ntf.BookCreated(ctx, req.Name, *req.Inventory)
+			log.Println(err)
```
cesarFuhr: Please test the error before logging.
You should only log if there was an error.
```suggestion
			if err != nil {
			        log.Println(err)
			}

File: cmd/api/main.go [hunk 5]
```diff
@@ -54,7 +57,34 @@ func run() error {
 		}
 	}
 
-	bookService := book.NewService(store)
+	//get Ntfy notifications config:
+	enableNotifications := true
```
cesarFuhr: The default behavior should be disabled Notifications.

cmd/api/main.go
```diff
@@ -54,7 +57,34 @@ func run() error {
...
+	enableNotificationsStr := os.Getenv("ENABLE_NOTIFICATIONS")
+	if enableNotificationsStr != "" {
+		enableNotifications, err = strconv.ParseBool(enableNotificationsStr)
+		if err != nil {
+			return fmt.Errorf("getting notifications enabler flag from env: %w", err)
+		}
+	}
```
cesarFuhr, line 68:
```suggestion
	}

File: cmd/api/notifications/notifications.go [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+	baseURL string
+	enabled bool
+	client  *http.Client
+}
+
+func NewNtfy(enableNotifications bool, notificationsBaseURL string) *Ntfy {
+	return &Ntfy{
+		baseURL: notificationsBaseURL,
+		enabled: enableNotifications,
+		client:  &http.Client{},
```
cesarFuhr: Please receive the `http.Client` as an argument.

cmd/api/notifications/notifications.go
```diff
@@ -0,0 +1,38 @@
...
+		if err != nil {
+			return fmt.Errorf("error delivering message ("+fmt.Sprintf("New book created: Title: %s Inventory: %v", title, inventory)+"): %w", err)
+		}
+		_, err = ntf.client.Do(req)
+		if err != nil {
+			return fmt.Errorf("error delivering message ("+fmt.Sprintf("New book created: Title: %s Inventory: %v", title, inventory)+"): %w", err)
+		}
+		return nil
+	}
+	return errors.New("notifications not enabled")
```
cesarFuhr: If its not enabled, its not enabled. Its not an error.
Change the logic here, the first few lines in this function should be:
```go
if !ntf.enabled {
        return nil
}

File: cmd/api/notifications/notifications.go (hunk 2 above)
Comment (cesarFuhr) [reply cmd/api/notifications/notifications.go:54]: This is the publisher of the message, you don't need to assert that the message was read anywhere.

File: cmd/api/notifications/notifications_test.go [hunk 9]
```diff
@@ -0,0 +1,84 @@
...
+
+		title := "book to test context timeout"
+		inventory := 40
+		notificationsTimeout := 2 * time.Millisecond
+		ctx, cancel := context.WithTimeout(context.Background(), notificationsTimeout)
+		defer cancel()
+
+		err := ntfy.BookCreated(ctx, title, inventory)
+		is.True(errors.Is(err, context.DeadlineExceeded))
+		is.Equal(err.Error(), `error delivering message (New book created: Title: book to test context timeout Inventory: 40): Post "https://ntfy.sh/test_Ah3mn6oD_New_book_created": context deadline exceeded`)
```
cesarFuhr: Don't test error messages like this.
Never assert error based on the string they return in the `Error()` method.

cesarFuhr: Create an error type or variable and return it.

cmd/api/notifications/notifications.go
```diff
@@ -0,0 +1,38 @@
...
+	return &Ntfy{
+		baseURL: notificationsBaseURL,
+		enabled: enableNotifications,
+		client:  &http.Client{},
+	}
+}
+
+func (ntf *Ntfy) BookCreated(ctx context.Context, title string, inventory int) error {
+	if ntf.enabled {
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, ntf.baseURL+"_New_book_created", strings.NewReader(fmt.Sprintf("New book created:\nTitle: %s\nInventory: %v", title, inventory))) //Ntfy SEEMS NOT TO ACEPT SLASHS OR DOTS AT TOPIC
```
cesarFuhr: Please break this down into more lines. There is too much stuff happening in a single line.
Create an `url` and a `message` variable to store the intermediary values.

cmd/api/notifications/notifications_test.go
```diff
@@ -0,0 +1,84 @@
...
+	notificationsBaseURL := "https://ntfy.sh/test_Ah3mn6oD"
+	enableNotifications := true
+	ntfy = NewNtfy(enableNotifications, notificationsBaseURL)
+
+	os.Exit(m.Run())
+}
+
+func TestBookCreated(t *testing.T) {
+
+	t.Run("notificates the criation of a new book without errors", func(t *testing.T) {
```
cesarFuhr: Its good to have a test that hits the actual service we are using, but please add a test that uses a mocked `http.Client` to assert the http request.
You could use a round tripper to assert the requests like this (https://www.reddit.com/r/golang/comments/gp53pu/using_roundtripper_to_mock_http_requests_in_go) or define a `Doer` interface, which would model the `client.Do` method (https://stackoverflow.com/questions/43240970/how-to-mock-http-client-do-method).

cmd/api/book/service.go
ramon-reichert: Off course! You're right.

cmd/api/main.go
ramon-reichert: Ok

cmd/api/notifications/notifications.go
ramon-reichert: ok

cmd/api/book/book_test.go
ramon-reichert: ok.

cmd/api/book/book_test.go
ramon-reichert: It was to be an error that I would try to catch at the log at stdout. But now that we decided not going so far, this error realy don't make sense anymore. I'll change the error to nil and assume it is a success.

cmd/api/notifications/notifications_test.go
ramon-reichert: Actually the error that realy matter for this test is the ´DeadlineExceeded`, tested on line 65.
Line 66 was just to see how ntfy returns the error, so I was intersted precisely on the string. It was already usefull, but not anymore. I think now we can take off this line 66. Can it be?

cmd/api/notifications/notifications.go
ramon-reichert: ok

cmd/api/book/book_test.go
```diff
@@ -39,13 +44,22 @@ func TestCreateBook(t *testing.T) {
 			return b, nil
 		})
 
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		mockNtfy.EXPECT().BookCreated(gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, _ book.Book) error {
```
cesarFuhr: Use `gomock.AssignableToTypeOf` to assert the type you received as the second argument.

cmd/api/notifications/notifications.go
```diff
@@ -0,0 +1,58 @@
...
+		return fmt.Errorf("error delivering message to ntfy (book ID: %v): %w", createdBook.ID, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		book.ErrStatusNotOK = book.ErrStatus{
+			StatusCode: resp.StatusCode,
+			Message:    "ntfy wrong response - want: 200 OK, got: " + resp.Status,
+		}
+		return book.ErrStatusNotOK
```
cesarFuhr: This is not the error type you are looking for.
You should have a `ErrNotificationFailed` or something similar. This is too specific.

cmd/api/notifications/notifications_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/books-service/cmd/api/book"
+	"github.com/books-service/cmd/api/notifications"
+	notificationmocks "github.com/books-service/cmd/api/notifications/mocks"
+	"github.com/google/uuid"
+	"github.com/matryer/is"
+	"go.uber.org/mock/gomock"
+)
+
+// Integration tests:
```
cesarFuhr: You already have an indicator of the test type in the name of the test function, I wouldn't add the comment, its pretty clear without it.

cmd/api/notifications/notifications_test.go
```diff
@@ -0,0 +1,176 @@
...
+		notificationsTimeout := 2 * time.Millisecond
+		ctx, cancel := context.WithTimeout(context.Background(), notificationsTimeout)
+		defer cancel()
+
+		err := ntfy.BookCreated(ctx, testerBook)
+		is.True(errors.Is(err, context.DeadlineExceeded))
+	})
+}
+
+// Unit tests:
```
cesarFuhr: Same here.

cmd/api/notifications/notifications_test.go
```diff
@@ -0,0 +1,176 @@
...
+		mockClient.EXPECT().Do(gomock.Any()).DoAndReturn(func(req *http.Request) (*http.Response, error) {
+			is.True(req.Method == http.MethodPost)
+			is.True(req.URL.String() == url)
+			requestedBody, _ := io.ReadAll(req.Body)
+			expectedBody, _ := io.ReadAll(message)
+			is.Equal(string(requestedBody), string(expectedBody))
+
+			resp := httptest.NewRecorder().Result()
+			resp.Status = "200 OK"
+			resp.StatusCode = 200
```
cesarFuhr: Use `http.StatusOK` instead.

cmd/api/notifications/notifications_test.go
```diff
@@ -0,0 +1,176 @@
...
+		mockClient.EXPECT().Do(gomock.Any()).DoAndReturn(func(req *http.Request) (*http.Response, error) {
+			is.True(req.Method == http.MethodPost)
+			is.True(req.URL.String() == url)
+			requestedBody, _ := io.ReadAll(req.Body)
+			expectedBody, _ := io.ReadAll(message)
+			is.Equal(string(requestedBody), string(expectedBody))
+
+			resp := httptest.NewRecorder().Result()
+			resp.Status = "500 Internal Server Error"
+			resp.StatusCode = 500
```
cesarFuhr: Here too, `http.StatusInternalServerError` or something like that.

cover.out
cesarFuhr, line 1: Don't commit this file.

ramon-reichert: 1- Should I add it to gitignore?
2 - Should I remove it from github directory?

ramon-reichert: ...

cesarFuhr: You should, remove, commit it and then add it to you add it to the .gitinore file (if you want to).

cesarFuhr (approved):

PR #33, opened 2024-02-13
implementing purchasing books feature

ramon-reichert:


ramon-reichert:
I'm a bit confused about WHERE to implement the orders methods...
Considering that:
- The requests/responses will be handled by the same http server that we are already using;
- The database will be the same too;
- We want isolation beetwen packages;
So, should we create a 'orders' package, conected with 'service' package through interfaces, or can we just implemente the new methods at the same 'service' package? Maybe just another file orders.go?

migrations/000005_new_tables.down.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS public.userstable;
```
cesarFuhr: Don't append table at the end of the table names.
`public.users` is enough.

migrations/000005_new_tables.up.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TYPE access AS ENUM ('user', 'admin');
+
+CREATE TABLE IF NOT EXISTS public.userstable
+(
+    user_id uuid PRIMARY KEY NOT NULL,
+    name text   NOT NULL,
+    user_access access DEFAULT 'user'
```
cesarFuhr: Maybe user `type`? or `role`? Role should be better.

migrations/000005_new_tables.up.sql
```diff
@@ -0,0 +1,37 @@
...
+
+CREATE TYPE order_status AS ENUM ('accepting_items', 'canceled', 'waiting_payment', 'paid');
+
+CREATE TABLE IF NOT EXISTS public.orderstable
+(
+order_id		uuid PRIMARY KEY NOT NULL,
+purchaser_id 	uuid,	
+order_status	order_status DEFAULT 'accepting_items',
+created_at 		timestamp with time zone DEFAULT now(),
+updated_at 	timestamp with time zone DEFAULT now()
```
cesarFuhr: What a awkward formatting. If you use spaces instead of tabs you wont have this issue.
Just don't get sucked in by the [holy wars](https://www.google.com/search?q=tabs+vs+spaces&sca_esv=297f8d6bee6ef7a9&source=hp&ei=MY3UZd_rJa7Z5OUP4pum4Aw&iflsig=ANes7DEAAAAAZdSbQfSrc1CoiH8SrSyNaZuq6m8s0szW&udm=&ved=0ahUKEwifyMm76LmEAxWuLLkGHeKNCcwQ4dUDCBU&uact=5&oq=tabs+vs+spaces&gs_lp=Egdnd3Mtd2l6Ig50YWJzIHZzIHNwYWNlczIFEAAYgAQyBhAAGBYYHjIGEAAYFhgeMgYQABgWGB4yBhAAGBYYHjIGEAAYFhgeMgYQABgWGB4yBhAAGBYYHjIGEAAYFhgeMgYQABgWGB5I0xlQAFi4GHABeACQAQCYAX6gAbUNqgEEMC4xNbgBA8gBAPgBAcICCxAAGIAEGLEDGIMBwgIREC4YgAQYsQMYgwEYxwEY0QPCAggQLhiABBixA8ICCBAAGIAEGLEDwgIIEC4YsQMYgATCAgcQABiABBgKwgILEC4YgAQYxwEYrwHCAgUQLhiABMICChAAGIAEGAoYsQPCAggQABgWGB4YCg&sclient=gws-wiz).

cesarFuhr:
>     * The requests/responses will be handled by the same http server that we are already using;

Yes and it has its own package `http`, which should use the inner packages to expose the API's.

>     * The database will be the same too;

Yes, but the database also has its own package. So that is OK too.

>     * We want isolation between packages;

We do, but sometimes the lines are blurred until we start implementing.

>       So, should we create a 'orders' package, connected with 'service' package through interfaces, or can we just implement the new methods at the same 'service' package? Maybe just another file orders.go?

Lets defer the decision of extracting it to another package until we need to. Start by implementing the orders functionality in the `book` package.

ramon-reichert:
Helloooo
Finally getting back to work!
I'm slowly refreshing my memory as I try to write the order functionality by TDD. Just a bit of code every day...

cmd/api/book/orders.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Order struct {
+	Order_ID     uuid.UUID
+	Purchaser_ID uuid.UUID
+	Order_status string
```
cesarFuhr: You are mixing snake_case and camelCase (or PascalCase), any particular reason to do that?
If not you stick to CamelCase since its the Go idiom.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,79 @@
...
+		Order_ID:     uuid.New(),
+		Purchaser_ID: user_id,
+		Order_status: "accepting_items",
+		CreatedAt:    createdAt,
+		UpdatedAt:    createdAt,
+	}
+	return s.repo.CreateOrder(ctx, newOrder)
+}
+
+type ItemAtOrder struct {
```
cesarFuhr:
```suggestion
type OrderItem struct {

File: cmd/api/book/orders.go [hunk 6]
```diff
@@ -0,0 +1,79 @@
...
+	Book_ID          uuid.UUID
+	Book_units       int
+	BookPriceAtOrder *float32
+	CreatedAt        time.Time
+	UpdatedAt        time.Time
+}
+
+type OrderItemsList struct {
+	Order     Order
+	ItemsList []ItemAtOrder
```
cesarFuhr:
```suggestion
	Items []OrderItems

File: cmd/api/book/orders.go [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+		Order:     order,
+		ItemsList: items,
+	}
+
+	return list, nil
+}
+
+type UpdateOrderRequest struct {
+	Order_ID   uuid.UUID
+	Book_ID    uuid.UUID
```
cesarFuhr: Same thing with the case here.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,79 @@
...
+type OrderItemsList struct {
+	Order     Order
+	ItemsList []ItemAtOrder
+}
+
+func (s *Service) ListOrderItems(ctx context.Context, order_id uuid.UUID) (OrderItemsList, error) {
+	order, items, err := s.repo.ListOrderItems(ctx, order_id)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return OrderItemsList{Order{}, []ItemAtOrder{}}, fmt.Errorf("timeout on call to ListOrderItems: %w", err)
```
cesarFuhr: This should be enough.
```suggestion
			return OrderItemsList{}, fmt.Errorf("timeout on call to ListOrderItems: %w", err)

File: cmd/api/book/orders_test.go [hunk 9]
```diff
@@ -0,0 +1,134 @@
...
+		ctrl := gomock.NewController(t)
+		mockRepo := bookmock.NewMockRepository(ctrl)
+		mockNtfy := bookmock.NewMockNotifier(ctrl)
+		mS := book.NewService(mockRepo, mockNtfy, notificationsTimeout)
+
+		someUser := uuid.New()
+
+		mockRepo.EXPECT().CreateOrder(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, o book.Order) (book.Order, error) {
+			is.True(o.OrderID != uuid.Nil)
+			is.True(o.PurchaserID == someUser)
```
cesarFuhr:
```suggestion
			is.Equal(o.PurchaserID, someUser)

Comment (cesarFuhr) [reply cmd/api/book/orders_test.go]: With `Equal` you can compare a full struct too, to avoid having issues with IDs you can set the expected to the actual before comparing.

File: cmd/api/database/database.go [hunk 10]
```diff
@@ -206,3 +207,185 @@ func (store *Store) ListBooksTotals(ctx context.Context, name string, minPrice32
...
+	var orderToReturn book.Order
+	err := createdRow.Scan(&orderToReturn.OrderID, &orderToReturn.PurchaserID, &orderToReturn.OrderStatus, &orderToReturn.CreatedAt, &orderToReturn.UpdatedAt)
+	if err != nil {
+		return book.Order{}, fmt.Errorf("storing order on db: %w", err)
+	}
+
+	return orderToReturn, nil
+}
+
+func (store *Store) ListOrderItems(ctx context.Context, order_id uuid.UUID) (book.Order, []book.OrderItem, error) {
```
ramon-reichert: I was thinking in put the []OrderItem as a field of Order, as you have said. Doing that, I could simplify other methods too,  always returning a full Order. Let me now what do you think...

cmd/api/book/orders.go
```diff
@@ -0,0 +1,81 @@
...
+	CreatedAt        time.Time
+	UpdatedAt        time.Time
+}
+
+type OrderItemsList struct {
+	Order     Order
+	ItemsList []OrderItem
+}
+
+func (s *Service) ListOrderItems(ctx context.Context, order_id uuid.UUID) (OrderItemsList, error) {
```
ramon-reichert: Here I would return the full Order, too. This method would be just a "reading" of the order, and maybe this name should change.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/google/uuid"
+)
+
+type Order struct {
+	OrderID     uuid.UUID
+	PurchaserID uuid.UUID
+	OrderStatus string
+	CreatedAt   time.Time
+	UpdatedAt   time.Time
```
ramon-reichert, line 19: Here I would add this field:
Items []OrderItem

cmd/api/book/orders.go
```diff
@@ -0,0 +1,81 @@
...
+		PurchaserID: user_id,
+		OrderStatus: "accepting_items",
+		CreatedAt:   createdAt,
+		UpdatedAt:   createdAt,
+	}
+	return s.repo.CreateOrder(ctx, newOrder)
+}
+
+type OrderItem struct {
+	OrderID          uuid.UUID
```
ramon-reichert: This OrderID would not be necessary again.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,81 @@
...
+	BookID           uuid.UUID
+	BookUnits        int
+	BookPriceAtOrder *float32
+	CreatedAt        time.Time
+	UpdatedAt        time.Time
+}
+
+type OrderItemsList struct {
```
ramon-reichert: And this struct would not be necessary anymore, too.

ramon-reichert:
I would like to improve this things before start to deal with the sql.Tx.
And maybe we could talk to define somethings that I've wrote in red here: https://docs.google.com/drawings/d/1fHuq1Q1CV_GkqYfjXuKSRbkUcMxCQ6KsOk5T1zNyUiY/edit

ramon-reichert (commented):
Take a look

cmd/api/database/database.go
cesarFuhr: I think is as good idea.

cesarFuhr: I think is as good idea.

cmd/api/book/orders.go
cesarFuhr: Agreed.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		errRepo := ErrResponse{
+			Code:    ErrResponseFromRespository.Code,
+			Message: ErrResponseFromRespository.Message + err.Error(),
+		}
+		return Order{}, errRepo
+	}
+
+	defer func() {
+		tx.Rollback()
```
cesarFuhr: You might want to at least log if any error comes back from this call.

ramon-reichert:
Hello! I think this feature is done. We also have the payment part to write, but it is just "more of te same", isn't it? I mean: I can write it, of course, but we could review this one here first and just keep this way later, right? Maybe, if you agree, we could even merge this branch already...
Now I will await your review, @cesarFuhr

cmd/api/book/orders.go
```diff
@@ -0,0 +1,161 @@
...
+	}()
+
+	err = txRepo.UpdateOrderRow(ctx, updtReq.OrderID) //changes field 'updated_at' and checks if the order is 'accepting_items'
+	if err != nil {
+		if errors.Is(err, ErrResponseOrderNotFound) {
+			return Order{}, ErrResponseOrderNotFound
+		} else if errors.Is(err, ErrResponseOrderNotAcceptingItems) {
+			return Order{}, ErrResponseOrderNotAcceptingItems
```
cesarFuhr: Why do you need this if you are returning exactly the same error?
You should be ok wrapping it.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,161 @@
...
+		return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
+	}
+
+	//Testing if there are sufficient inventory of the book asked, and if is not archived:
+	bk, err := txRepo.GetBookByID(ctx, updtReq.BookID)
+	if err != nil {
+		if errors.Is(err, ErrResponseBookNotFound) {
+			return Order{}, ErrResponseBookNotFound
+		}
+		return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
```
cesarFuhr: Usually we would put some more context in this wrap, something like:
```
fetching book from database: %w

File: cmd/api/book/orders.go [hunk 18]
```diff
@@ -0,0 +1,161 @@
...
+		return Order{}, ErrResponseInsufficientInventory
+	}
+
+	//Testing if the book is already at the order and, if it is, getting it:
+	bookAtOrder, err := txRepo.GetOrderItem(ctx, updtReq.OrderID, updtReq.BookID)
+	if err != nil {
+		if !errors.Is(err, ErrResponseBookNotAtOrder) {
+			return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
+		}
+	}
```
cesarFuhr: The outer `if err != nil` is not needed here.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,161 @@
...
+		//Created_at and Updated_at fields will be set properly at database layer
+
+		bookAtOrder, err = txRepo.UpsertOrderItem(ctx, updtReq.OrderID, bookAtOrder)
+		if err != nil {
+			return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
+		}
+
+	} else { //Case the book is already at the order, and book_units becomes zero from update, the book is excluded from the order. Even so, it must be updated at bookstable.
+
+		if errors.Is(err, ErrResponseBookNotAtOrder) { //But, if the book is not at order, a request attempting to decrease its units value can mean an error from client, so an error is returned.
```
cesarFuhr: This is a bit confusing, this error check so far away from where the error was originated.
You should, at least, name the error better (with more detail) so we know what we are talking about after this many lines.

cmd/api/book/orders.go
```diff
@@ -0,0 +1,161 @@
...
+	bk.UpdatedAt = time.Now().UTC().Round(time.Millisecond)
+	_, err = txRepo.UpdateBook(ctx, bk)
+	if err != nil {
+		return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
+	}
+
+	updatedOrder, err := txRepo.ListOrderItems(ctx, updtReq.OrderID)
+	if err != nil {
+		return Order{}, fmt.Errorf("error on call to UpdateOrderTx: %w ", err)
+	}
```
cesarFuhr: This List call can be outside the transaction.

cmd/api/http/http_test.go
```diff
@@ -334,3 +334,7 @@ func pagedBooksToResponse(page book.PagedBooks) PageOfBooksResponse {
 		Results:     results,
 	}
 }
+
+//func TestCreateOrder(t *testing.T) { TO DO }
+
+//func TestUpdateOrder(t *testing.T) { TO DO }
```
cesarFuhr: Its pretty common to leave `//TODO: what needs to be done in the code` and people will search for it when they have some extra time to fix something small.

cesarFuhr (approved):

PR #35, opened 2025-02-20
Feat/in memory storage

ramon-reichert:
Implementation of an in-memory storage using the memDb library as an alternative to the Postgres database, accessed through the same Repository interface.
The challenge was to keep Repository signatures untouched and make it work somehow.
I used ChatGPT to help me understand the concept of in-memory storage and to give me options on how to implement it.
I thought it was great because AI gave me the agility to speculate about different approaches before actually starting to write code.
It can also help with many manual tasks.
Of course, I had to understand and adapt the generated pieces of code, but I found it very helpful.
I've already tested the code with Postman and written unit tests. I think the next step is to make it flexible in choosing which storage to use by setting some environment variables.


ramon-reichert:
It seems to be working even with two docker-compose files, and I know how to manage them in my machine, but I'm not sure how it is working here inside github and fly. Can you help me to clarify it?

